PR #23, opened 2019-11-21
Run stylish-haskell in all files! 💅🏼

kutyel:
The moment has arrived!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -1,0 +1,63 @@
+steps:
+  - simple_align:
+      cases: true
+      top_level_patterns: true
+      records: true
+
+  # Import cleanup
+  - imports:
+      align: global
+      list_align: after_alias
+      pad_module_names: true
+      long_list_align: inline
+      empty_list_align: inherit
+      list_padding: 4
+      separate_lists: true
+      space_surround: false
+
+  # Language pragmas
+  - language_pragmas:
+      style: vertical
+      align: true
+      remove_redundant: true
+      language_prefix: language
+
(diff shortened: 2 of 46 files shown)


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# editorconfig.org
```
serras: Why do we need this file?

.stylish-haskell.yaml
```diff
@@ -0,0 +1,63 @@
+steps:
```
serras: Why is this file so big? I thought we were just using the defaults except for the `language_prefix`.

.stylish-haskell.yaml
```diff
@@ -0,0 +1,63 @@
...
+      remove_redundant: true
+      language_prefix: language
+
+  # Remove trailing whitespace
+  - trailing_whitespace: {}
+
+columns: 100
+newline: native
+cabal: true
+language_extensions:
```
serras: Why is there a list of language extensions here?

stack-nightly.yaml
```diff
@@ -21,3 +21,5 @@ extra-deps:
 - http2-client-grpc-0.8.0.0
 - avro-0.4.5.4
 - language-protobuf-1.0
+- github: jaspervdj/stylish-haskell
```
serras: These should be removed. Instead, please write a `development.md` file in which you explain that we are using an unreleased branch of `stylish-haskell`.

stack.yaml
```diff
@@ -20,6 +20,8 @@ extra-deps:
 - http2-client-grpc-0.8.0.0
 - avro-0.4.5.4
 - language-protobuf-1.0
+- github: jaspervdj/stylish-haskell
```
serras: To be removed

.editorconfig
kutyel: https://editorconfig.org/ 😄

.stylish-haskell.yaml
kutyel: Apparently you need to provide the whole YAML file from which stylish-haskell needs to read 😞

.stylish-haskell.yaml
kutyel: This is for the parser, AFAIK

PR #24, opened 2019-11-22
Add example todolist ✅

### LINKED ISSUE #10: Port examples from Scala Mu
The Mu repo has [quite some examples](https://github.com/higherkindness/mu/tree/master/modules/examples). We should port them to `mu-haskell`.

- [x] health-check
- [x] route-guide
- [x] todolist
- [x] seed

kutyel:
partially fixes #10


examples/todolist/src/Server.hs
```diff
@@ -0,0 +1,85 @@
...
+  putStr "insert: " >> print (titl, tgId)
+  atomically $ do
+    modifyTVar oldId (+1)
+    newId <- readTVar oldId
+    let newTodo = TodoListMessage newId tgId titl
+    modifyTVar t (newTodo:)
+    pure $ TodoListResponse newTodo
+
+getMsg :: Int32 -> TodoListMessage -> Bool
+getMsg val (TodoListMessage idM _ _) = idM == val
```
serras: Since `TodoListMessage` is a record, you might consider using the field accessor instead.

examples/todolist/src/Server.hs
```diff
@@ -0,0 +1,85 @@
...
+
+retrieve :: TodoList -> MessageId -> IO TodoListResponse
+retrieve t (MessageId idMsg) = do
+  putStr "retrieve: " >> print idMsg
+  atomically $ do
+    todos <- readTVar t
+    pure $ TodoListResponse $ fromMaybe
```
serras: Try to limit the amount of computation you do in an `atomically` block. I would recomend:

```haskell
todos <- readTVarIO t
let todo = fromMaybe ...
return $ TodoListReponse todo
```

Yes, there's a specific `readTVarIO` which is just `atomically . readTVar` :P

examples/todolist/src/Server.hs
```diff
@@ -0,0 +1,85 @@
...
+  atomically $ do
+    todos <- readTVar t
+    pure $ TodoListList todos
+
+update :: TodoList -> TodoListMessage -> IO TodoListResponse
+update t mg@(TodoListMessage idM titM tgM) = do
+  putStr "update: " >> print (idM, titM, tgM)
+  atomically $ do
+    modifyTVar t (fmap (\m -> if getMsg idM m then mg else m))
+    pure $ TodoListResponse mg
```
serras: Same as above, do not do many things in `atomically`:

```haskell
putStr ...
atomically $ modifyTVar ...
pure $ TodoListResponde mg
```

examples/todolist/src/Server.hs
```diff
@@ -0,0 +1,85 @@
...
+
+destroy :: TodoList -> MessageId -> IO MessageId
+destroy t (MessageId idMsg) = do
+  putStr "destroy: " >> print idMsg
+  atomically $ do
+    todos <- readTVar t
+    case find (getMsg idMsg) todos of
```
serras: This one I leave it for you!

examples/todolist/src/Server.hs
```diff
@@ -67,19 +69,17 @@ list_ t = do
...
-    todos <- readTVar t
-    case find (getMsg idMsg) todos of
-      Just todo -> do
-        modifyTVar t (filter (/=todo))
-        pure $ MessageId idMsg -- OK ✅
-      Nothing   -> pure $ MessageId 0 -- did nothing
+  todos <- readTVarIO t
+  case find (getMsg idMsg) todos of
+    Just todo -> do
+      atomically $ modifyTVar t (filter (/=todo))
```
serras: Now I see this is wrong! `t` could change in between calls, so everything must be in the same `atomically` block.

PR #25, opened 2019-11-25
Add API for exceptional error conditions

serras:
Following a meeting with @kutyel and @pepegar, we have drafted an API for exceptional conditions in `mu-haskell` (errors like *not found* or *not implemented*).

* We follow the design of both Servant and Mu, that is, and make the servers return `ExceptT ServerError IO a` instead of plain `IO a` as before.
* To make simple server unaware of the extra `ExceptT` layer, there's an `alwaysOk` function which lifts a simple handler in `IO` to work on `ExceptT ServerError IO`.

This PR implements that API and updates the examples to use it.

Comments are welcome. The main paint points are: 
* Conduits also have to be upgraded to use those monad transformers (do they? maybe we can make them unaware of external errors, but this does not seam real...)
* You have to sprinkle the `alwaysOk` here and there.

### DIFF
--- a/core/rpc/src/Mu/Server.hs
+++ b/core/rpc/src/Mu/Server.hs
@@ -29,24 +29,50 @@ module Mu.Server (
   -- * Servers and handlers
   ServerIO, ServerT(..)
 , HandlersIO, HandlersT(..)
+  -- * Errors which might be raised
+, serverError, ServerErrorIO, ServerError(..), ServerErrorCode(..)
+  -- ** Useful when you do not want to deal with errors
+, alwaysOk
 ) where
 
+import           Control.Monad.Except
 import           Data.Conduit
 import           Data.Kind
 
 import           Mu.Rpc
 import           Mu.Schema
 
+serverError :: ServerError -> ServerErrorIO a
+serverError = throwError
+
+alwaysOk :: IO a -> ServerErrorIO a
+alwaysOk = liftIO
+
+data ServerError
+  = ServerError ServerErrorCode String
+
+data ServerErrorCode
+  = Unknown
+  | Unavailable
+  | Unimplemented
+  | Unauthenticated
+  | Internal
+  | Invalid
+  | NotFound
+  deriving (Eq, Show)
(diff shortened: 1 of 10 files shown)


core/rpc/src/Mu/Server.hs
```diff
@@ -60,13 +86,14 @@ instance HandlesRef ('FromRegistry subject t last) t
 -- Arguments
 instance (HandlesRef ref t, Handles args ret m h, handler ~ (t -> h))
          => Handles ('ArgSingle ref ': args) ret m handler
-instance (HandlesRef ref t, Handles args ret m h, handler ~ (ConduitT () t IO () -> h))
+instance (HandlesRef ref t, Handles args ret m h, handler ~ (ConduitT () t m () -> h))
```
pepegar: Now that we're handling it with `ServerError`... Should we keep this instance? this might be a scapehatch for using monads that don't have a `MonadError ServerError`

serras: Indeed, this was wrong. I just pushed an update to require `m` to be a `MonadError`.

examples/todolist/src/Server.hs
```diff
@@ -54,28 +54,29 @@ insert oldId t (TodoListRequest titl tgId) = do
...
 retrieve t (MessageId idMsg) = do
-  putStr "retrieve: " >> print idMsg
-  todos <- readTVarIO t
-  let todo = fromMaybe (TodoListMessage 0 0 "I don't know" False) (find (getMsg idMsg) todos)
-  pure $ TodoListResponse todo -- FIXME: what if it's not found?
+  liftIO (putStr "retrieve: " >> print idMsg)
+  todos <- liftIO $ readTVarIO t
+  case find (getMsg idMsg) todos of
+    Just todo -> pure $ TodoListResponse todo
+    Nothing   -> serverError $ ServerError NotFound "unknown todolist id"
```
kutyel: 👀 This looks good, thanks for adding it! 👏

PR #27, opened 2019-12-03
Add Seed example 🌱

### LINKED ISSUE #10: Port examples from Scala Mu

kutyel:
Closes #10


docs/intro.md
```diff
@@ -0,0 +1,94 @@
+# Introduction to Mu-Haskell
+
+## What is Mu-Haskell?
```
diesalbla: I would suggest a phrase describing the context, the trend, and the problem or difficulty which Mu-Haskell is intended to solve. 

Here is some sketches, fill as convenient: 

> Many companies have embraced Micro-Services architectures as the best way to scale up their internal software systems, separate work across different company divisions and development teams. 
> Microservices architectures also allow teams to turn an idea or bug report into a working feature of fix in production more quickly, in accordance to the agile principles.   
> However, micro-services are not without costs. Every connection between micro-services becomes now a boundary that requires one to act as a server, and the other to act as the client. Each part implementation needs to add the protocol, the codification of the data for transmission, etc.   
> Also, business logic of the application starts to spread around several code-bases, making it difficult to maintain.

kutyel: Hi Diego! Thanks, you are right, thanks for the recommendation! Unfortunately, this change goes on this https://github.com/higherkindness/mu-haskell/pull/26PR from @serras, I started this branch from his new docs branch to dogfood the new Stack template he created (templates/grpc-server.hsfiles) 😄

examples/seed/stack.yaml
```diff
@@ -0,0 +1,14 @@
+resolver: lts-14.16
```
serras: You don't need this file if you have another `stack.yaml` in the root.
I know this has been auto-generated, but it's better to remove it to not confuse casual readers.

PR #29, opened 2019-12-03
Separate annotations from core schema type

serras:
Before definitions like the id for ProtoBuf were part of the schema. However, that meant that we could not retro-fit the same schema for different purposes. This PR separates these two concepts and updates the Template Haskell code, so examples don't need to change.

This change will be properly documented in #31

### DIFF
--- a/adapter/avro/src/Mu/Adapter/Avro.hs
+++ b/adapter/avro/src/Mu/Adapter/Avro.hs
@@ -94,12 +94,12 @@ instance forall r d ds.
 -- HasAvroSchema instances
 
 instance (KnownName name, HasAvroSchemaFields sch args)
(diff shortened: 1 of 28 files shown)


serras:
The Travis failure is so weird. How can it fail in the PR but not in the branch? Furthermore, it seems completely non-reproducible in my machine, with either LTS or Nightly.

serras:
Now I understand, the problem is that it's conflicting with the `seed` example, which is not yet merge.
@kutyel could you try adding `{-# language TypeFamilies #-}` to your Definition file in that example, and then run the build again? Thanks 🙇

kutyel:
Apparently it didn't work 😖

serras:
I am fairly confident that now it should work. Please review and merge if it's green :)

PR #34, opened 2019-12-04
Add recommended vscode extensions! 💻

kutyel:
Apparently this is a new feature of **vscode**, and when you open a project with this settings a prompt appears asking you to auto-install the recommended extensions! 👏  I think this behaviour is quite handy and will help the new contributors to follow https://github.com/higherkindness/mu-haskell/blob/master/DEVELOPMENT.md 😉

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": true
+}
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
serras: I love the idea of the recommendations, but I am not 100% sure about this file. The `settings.json` *mandates* this setting to be on, but people may not like these things to be forced on them.

I know some projects use Git hooks to ensure that styling is consistent. Maybe we can look at that? Although I think it's not really worth it.

PR #35, opened 2019-12-05
Add support for logging and other transformers

### LINKED ISSUE #32: Add support for logging
Make `ServerErrorIO` in `rpc` include a new layer of `LoggingT` from [`monad-logger`](http://hackage.haskell.org/package/monad-logger).

serras:
Closes #32

### DIFF
--- a/core/rpc/src/Mu/Rpc/Examples.hs
+++ b/core/rpc/src/Mu/Rpc/Examples.hs
@@ -1,6 +1,7 @@
 {-# language DataKinds             #-}
 {-# language DeriveAnyClass        #-}
 {-# language DeriveGeneric         #-}
+{-# language FlexibleContexts      #-}
 {-# language FlexibleInstances     #-}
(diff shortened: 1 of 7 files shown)


core/rpc/src/Mu/Rpc/Examples.hs
```diff
@@ -9,7 +10,7 @@
 {-# language PolyKinds             #-}
 {-# language TypeFamilies          #-}
 {-# language TypeOperators         #-}
-{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
 module Mu.Rpc.Examples where
```
kutyel: Maybe we should move this example to the `examples` folder? 🤔

serras: I like to have them here as a small examples people can see while browsing the API docs.

PR #37, opened 2019-12-06
Work on docs

### LINKED ISSUE #31: Document new annotations design
Some changes need to be made to mu-schema to support in the future things like having parametrised queries in GraphQL!

### LINKED ISSUE #36: Document support for transformers in ServerT
In #35, we not only introduced logging but also generalised it to any monad transformer! 💯 

Now, of course, we need to document and showcase this 😄

serras:
Fixes #31 
Fixes #36

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -52,7 +52,7 @@ main = do
   run 8080 $ metrics waiMetrics (gRpcApp server)
 ```
 
-Another possibility is to expose [Prometheus](https://prometheus.io/) metrics. In this case, a specific monitoring process scrapes your servers from time to time. The gathered data can later be analyzed and visualized. Luckily, there's some [middlaware](https://github.com/fimad/prometheus-haskell) exposing the require endpoints automatically.
+Another possibility is to expose [Prometheus](https://prometheus.io/) metrics. In this case, a specific monitoring process scrapes your servers from time to time. The gathered data can later be analyzed and visualized. Luckily, there's some [middleware](https://github.com/fimad/prometheus-haskell) exposing the require endpoints automatically.
 
 ```haskell
 import Network.Wai.Handler.Warp (run)
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -1,1 +1,68 @@
 # Integration with WAI middleware
+
+Although you usually run a `mu-rpc` server directly using a function like `runGRpcApp`, this is just a convenience function to make it simpler to run the server. Under the hood, the library generates a so-called WAI application, which is then fed to an actual server.
+
+WAI stands for [*Web Application Interface*](https://www.yesodweb.com/book/web-application-interface). WAI defines a common set of APIs against which web application can be developed. Web servers, like [Warp](http://www.aosabook.org/en/posa/warp.html), take an application which uses that API and serves it against the actual network. The main benefit of this separation is that applications and servers can evolve separately, without a tight coupling between them.
+
+One of the features that WAI provides is the definition of *middleware* components. *Middlewares* are not complete applications; instead they wrap another application to provide additional capabilities. There is a whole ecosystem of WAI middleware, as one can notice by [searching `wai middleware` in Hackage](http://hackage.haskell.org/packages/search?terms=wai+middleware).
+
+## Serving static files
+
+It's a common task in web servers to send some static content for a subset of URLs (think of resources such as images or JavaScript code). `wai-middleware-static` automates that task, and also serves as the simplest example of WAI middleware.
+
+Remember that in our [original code](intro.md) our `main` function looked as follows:
+
+```haskell
+main = runGRpcApp 8080 server
+```
+
+We can split this single line into two phases: first creating the WAI application by means of `gRpcApp`, and then running it using Warp's `run` function.
+
+```haskell
+import Network.Wai.Handler.Warp (run)
+
+main = run 8080 $ gRpcApp server
+```
+
+Right in the middle of the two steps we can inject any middleware we want.
+
+```haskell
+import Network.Wai.Handler.Warp (run)
+import Network.Wai.Middleware.Static
+
+main = run 8080 $ static (gRpcApp server)
+```
+
+With that simple change, our web server now first checks whether a file with the requested name exists in the directory from which the application is running. If that is the case, it's cached and served, otherwise the underlying gRPC application is run. Needless to say, this behavior might not be the desired one, so the library provides [`staticPolicy`](http://hackage.haskell.org/package/wai-middleware-static/docs/Network-Wai-Middleware-Static.html#v:staticPolicy) to customize it.
+
+## Metrics
+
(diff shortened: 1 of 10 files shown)


docs/middleware.md
```diff
@@ -1 +1,68 @@
...
+
+main = do
+  -- Taken from the official documentation
+  store <- serverMetricStore <$> forkServer "localhost" 8000
+  waiMetrics <- registerWaiMetrics store
+  -- Wrap the application and run it
+  run 8080 $ metrics waiMetrics (gRpcApp server)
+```
+
+Another possibility is to expose [Prometheus](https://prometheus.io/) metrics. In this case, a specific monitoring process scrapes your servers from time to time. The gathered data can later be analyzed and visualized. Luckily, there's some [middlaware](https://github.com/fimad/prometheus-haskell) exposing the require endpoints automatically.
```
kutyel:
```suggestion
Another possibility is to expose [Prometheus](https://prometheus.io/) metrics. In this case, a specific monitoring process scrapes your servers from time to time. The gathered data can later be analyzed and visualized. Luckily, there's some [middleware](https://github.com/fimad/prometheus-haskell) exposing the require endpoints automatically.
```

docs/rpc.md
```diff
@@ -86,15 +91,22 @@ Note that depending on the concrete implementation you use to run the server, on
...
 sayHello (HelloRequest nm) = return (HelloResponse ("hi, " <> nm))
 ```
 
-Since you can declare more than one method in a service, you need to join then into a `Server`. You do so by using `(:<|>:)` between each handler and ending the sequence with `H0`. In addition to the name of the service, `Server` has an additional parameter which records the types of the handlers. Since that list may become quite long, we can ask GHC to write it for us by using the `PartialTypeSignatures` extension and writing an underscore `_` in that position. One final observation is that in the code below we are using `ServerIO`, which is an instance of `Server` which allows running `IO` operations.
+Notice the use of `MonadServer` in this case. This gives us the ability to:
+
+* Run arbitrary `IO` actions by using `liftIO`,
+* Return an error code by calling `serverError`.
+
+Being polymorphic here allows us to run the same server in multiple back-ends. Furthermore, by enlargint the set of abilities required for our monad `m`, we can [integrate with other libraries](transformer.md), including logging and resource pools.
```
kutyel:
```suggestion
Being polymorphic here allows us to run the same server in multiple back-ends. Furthermore, by enlarging the set of abilities required for our monad `m`, we can [integrate with other libraries](transformer.md), including logging and resource pools.
```

docs/rpc.md
```diff
@@ -86,15 +91,22 @@ Note that depending on the concrete implementation you use to run the server, on
...
+
+Since you can declare more than one method in a service, you need to join then into a `Server`. You do so by using `(:<|>:)` between each handler and ending the sequence with `H0`. In addition to the name of the service, `Server` has an additional parameter which records the types of the handlers. Since that list may become quite long, we can ask GHC to write it for us by using the `PartialTypeSignatures` extension and writing an underscore `_` in that position.
```
kutyel:
```suggestion
Since you can declare more than one method in a service, you need to join them into a `Server`. You do so by using `(:<|>:)` between each handler and ending the sequence with `H0`. In addition to the name of the service, `Server` has an additional parameter which records the types of the handlers. Since that list may become quite long, we can ask GHC to write it for us by using the `PartialTypeSignatures` extension and writing an underscore `_` in that position.
```

docs/schema.md
```diff
@@ -48,45 +48,26 @@ Note that GHC requires all of `DEnum`, `DRecord`, `FieldDef`, and so forth to be
...
+The idea is that your schema lives in an external file, so you can share it with other components of your system. To declare that we want the file to be pre-processed before compilation, we use a GHC feature called `TemplateHaskell`, hence the initial line starting with `language`.
 
 ```haskell
-{-# language QuasiQuotes #-}
+{-# language TemplateHaskell #-}
 
-type ExampleSchema = [protobufFile|path/to/file.proto|]
+import Mu.Quasi.ProtoBuf
+
+protobuf "ExampleSchema" "path/to/file.proto"
```
kutyel: Here is `protobuf` or `protobufFile`? 🤔 I didn't know you could do this without QuasiQuotes! 💥

docs/transformer.md
```diff
@@ -0,0 +1,84 @@
...
+
+The most important addition with respect to the [original code](rpc.md) is in the signature. Before we only had `MonadServer m`, now we have an additional `MonadLogger m` there.
+
+As we have done with the Reader example, we need to define how to handle `MonadLogger`. `monad-logger` provides [three different monad transformers](http://hackage.haskell.org/package/monad-logger-0.3.31/docs/Control-Monad-Logger.html#g:3), so you can choose whether your logging will be completely ignored, will become a Haskell value, or would fire some `IO` action like printing in the console. Each of these monad transformers comes with a `run` action which declares how to handle it; the extended function `runGRpcAppTrans` takes that handler as argument.
+
+```haskell
+main = runGRpcAppTrans 8080 runStderrLoggingT quickstartServer
+```
+
+If you prefer other logging library, this is fine with us! Replacing `monad-logger` by [`co-log`](https://github.com/kowainik/co-log) means asking for a different capability in the server. In this case we have to declare the type of the log items as part of the `WithLog` constraint:
```
kutyel: I think @chshersh and @vrom911 are gonna love this! ❤️

docs/transformer.md
```diff
@@ -0,0 +1,84 @@
...
+
+```haskell
+import Colog.Monad
+
+sayHello :: (MonadServer m, WithLog env String m)
+         => HelloRequest -> m HelloResponse
+sayHello (HelloRequest nm) = do
+  logInfoN "running hi"
```
kutyel: I think this needs to be `logDebug` or something like that from the `co-log` library? 🤔

docs/transformer.md
```diff
@@ -0,0 +1,84 @@
...
+```haskell
+main = runGRpcAppTrans 8080 logger quickstartServer
+  where logger = usingLoggerT (LogAction $ liftIO putStrLn)
+```
+
+## Warning
+
+The `run` function you provide to `runGRpcAppTrans` may be called more than once! This is fine for readers and logging, but not for `StateT`, for example. In particular, you must ensure that your `run` function is *idempotent*, that is, that the result of calling it more than once is the same as calling it just once.
+
+In the particular case of `StateT`, we suggest using a [transactional variable](http://hackage.haskell.org/package/stm/docs/Control-Concurrent-STM-TVar.html), passed as either an argument or uisng `ReaderT`. This has the additional benefit that concurrent access to the variable - which is fairly possible in a gRPC server -- are automatically protected for data races and deadlocks.
```
kutyel:
```suggestion
In the particular case of `StateT`, we suggest using a [transactional variable](http://hackage.haskell.org/package/stm/docs/Control-Concurrent-STM-TVar.html), passed as either an argument or using `ReaderT`. This has the additional benefit that concurrent access to the variable - which is fairly possible in a gRPC server -- are automatically protected for data races and deadlocks.
```

docs/schema.md
serras: One of the changes due to the new annotation format is that everything now uses `TemplateHaskell`, quasi-quotation is no longer supported.

PR #40, opened 2019-12-11
Implement Persistent example 💾

### LINKED ISSUE #30: Example with Persistent
The idea is to have a simple gRPC server which shows how one can integrate our library with Persistent for database queries. The server should roughly provide the following functions (the signatures are pseudo-proto):

```scala
rpc getPerson(id: PersonId) returns Person
rpc newPerson(p: Person) returns PersonId
rpc allPeople() returns stream Person
```

I suggest to use [`persistent-sqlite`](http://hackage.haskell.org/package/persistent-sqlite) as the concrete realization of the DB layer, but keep the code polymorphic over the choice of DB as much as possible.

kutyel:
Fixes #30

### DIFF
--- a/examples/with-persistent/mu-example-with-persistent.cabal
+++ b/examples/with-persistent/mu-example-with-persistent.cabal
@@ -13,7 +13,7 @@ extra-source-files:  README.md
 
 executable persistent-server
   hs-source-dirs:      src
-  main-is:             Main.hs
+  main-is:             Server.hs
   other-modules:       Schema
   default-language:    Haskell2010
   build-depends:       base >= 4.12 && < 5
--- a/examples/with-persistent/mu-example-with-persistent.cabal
+++ b/examples/with-persistent/mu-example-with-persistent.cabal
@@ -1,0 +1,28 @@
+name:                mu-example-with-persistent
+version:             0.1.0.0
+-- synopsis:
+-- description:
+homepage:            https://github.com/higherkindness/mu-haskell/examples/with-persistent#readme
+author:              Flavio Corpa
+maintainer:          flavio.corpa@47deg.com
+copyright:           Copyright © 2019-2020 47 Degrees. <http://47deg.com>
+category:            Network
+build-type:          Simple
+cabal-version:       >=1.10
+extra-source-files:  README.md
+
+executable persistent-server
+  hs-source-dirs:      src
+  main-is:             Main.hs
+  other-modules:       Schema
+  default-language:    Haskell2010
+  build-depends:       base >= 4.12 && < 5
+                     , conduit
+                     , mu-schema
+                     , mu-rpc
+                     , mu-protobuf
+                     , mu-grpc-server
+                     , persistent
+                     , persistent-sqlite
+                     , persistent-template
+                     , text
--- a/examples/with-persistent/mu-example-with-persistent.cabal
+++ b/examples/with-persistent/mu-example-with-persistent.cabal
@@ -26,5 +26,4 @@ executable persistent-server
                      , persistent
                      , persistent-sqlite
                      , persistent-template
-                     , resource-pool
                      , text
--- a/examples/with-persistent/mu-example-with-persistent.cabal
+++ b/examples/with-persistent/mu-example-with-persistent.cabal
@@ -18,6 +18,7 @@ executable persistent-server
   default-language:    Haskell2010
   build-depends:       base >= 4.12 && < 5
                      , conduit
+                     , monad-logger
                      , mu-schema
                      , mu-rpc
                      , mu-protobuf
@@ -25,4 +26,5 @@ executable persistent-server
                      , persistent
                      , persistent-sqlite
                      , persistent-template
+                     , resource-pool
                      , text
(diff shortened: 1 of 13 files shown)


examples/with-persistent/src/Main.hs
```diff
@@ -0,0 +1,43 @@
...
+import           Mu.GRpc.Server
+import           Mu.Server
+
+import           Schema
+
+main :: IO ()
+main = do
+  putStrLn "running application with persistent"
+  runMigration migrateAll
+  runNoLoggingT $
```
serras: It would be great if you could use logging, even if it's only stdout logging.

examples/with-persistent/src/Main.hs
```diff
@@ -0,0 +1,43 @@
...
+    withSqlitePool @(NoLoggingT IO) @SqlBackend ":memory:" 10 $ \pool ->
+      runGRpcApp 1234 server pool
+
+server :: MonadServer m => Pool SqlBackend -> ServerT PersistentService m _
+server p = Server (getPerson p :<|>: newPerson p :<|>: allPeople p :<|>: H0)
+
+getPerson :: Monad m => Pool SqlBackend -> PersonRequest -> m Person
+getPerson pool (PersonRequest name) = liftIO $ flip runSqlPersistMPool pool $ do
+  p <- getBy $ UniquePerson name
```
serras: Like in the case of `STM` blocks, it's important to bound the amount of code run in a `runSqlPersistMPool` block, because during that time we are taking the connection for our use. In this case the block is small enough, but be aware of that fact.

examples/with-persistent/src/Schema.hs
```diff
@@ -0,0 +1,45 @@
...
+share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
+Person json
+  name         String
+  age          Int
+  UniquePerson name
+  deriving Show
+|]
+
+deriving instance Generic Person
+deriving instance HasSchema PersistentSchema "Person" Person
```
serras: How did you get this to work? Or you still need to add the hand-written instance here?

kutyel: This is working now in the last commit! 🙌

examples/with-persistent/src/Main.hs
```diff
@@ -0,0 +1,58 @@
...
+
+newPerson :: (MonadServer m, MonadLogger m)
+          => Pool SqlBackend
+          -> Person
+          -> m PersonRequest
+newPerson pool p@(Person name _) = liftIO $ flip runSqlPersistMPool pool $ do
+  logDebugN $ T.pack $ "inserting person: " ++ show p
+  newId <- insert p
+  pure $ case newId of
+    PersonKey (SqlBackendKey nId) -> PersonRequest nId
```
serras: Could this be made better by using `fmap`? I am thinking of something along the lines of

```haskell
PersonRequest . unPersonKey . unSqlBackendKey <$> insert p
```

examples/with-persistent/src/Main.hs
```diff
@@ -0,0 +1,58 @@
...
+
+allPeople :: (MonadServer m, MonadLogger m)
+          => Pool SqlBackend
+          -> ConduitT Person Void m ()
+          -> m ()
+allPeople pool = liftIO $ flip runSqlPersistMPool pool $ do
+  logDebugN $ T.pack "streaming people..."
+  pure $ selectSource [] []
```
serras: This doesn't do anything. You need to connect the source you obtain from `selectSource` with the sink. Or am I missing something and your code does this in some complicated way? I was expecting something along the lines of:

```haskell
runConduit $ selectSource [] [] .| sink
```

examples/with-persistent/src/Schema.hs
```diff
@@ -0,0 +1,49 @@
...
+share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
+Person json
+  name T.Text
+  age  Int32
+  deriving Show
+|]
+
+deriving instance Generic Person
+
+-- Unfortunately we need to write this instance by hand 😔 (for now!)
```
serras: In fact, it's quite possible to write a generic instance on the lines of:

```haskell
instance Generic t => HasSchema sch sty (Entity t)
```

We just need to "reuse" the `HasSchema` for `t`, and then assume that there's an `id` field which is mapped to the corresponding field in the `Entity`. Do you want to give it a try?

examples/with-persistent/mu-example-with-persistent.cabal
```diff
@@ -0,0 +1,30 @@
...
+homepage:            https://github.com/higherkindness/mu-haskell/examples/with-persistent#readme
+author:              Flavio Corpa
+maintainer:          flavio.corpa@47deg.com
+copyright:           Copyright © 2019-2020 47 Degrees. <http://47deg.com>
+category:            Network
+build-type:          Simple
+cabal-version:       >=1.10
+extra-source-files:  README.md
+
+executable persistent-server
```
serras: I would be awesome if you could also create a client for this example. That way, we can tell people that they can run the server, and then make queries against it using the client.

serras: The client does not involve any DB, so it should be fairly easy to write.

PR #41, opened 2019-12-11
[WIP] Higher-kinded data

### LINKED ISSUE #39: [Design] Use higher-kinded data
Inspired by some of the issues @cb372 raised about Proto3 and their "everything is optional" policy, I have been thinking of a new design to accomodate this scenario, but also others like GraphQL in which sometimes you want to use a data type only to indicate that you want to query some field.

Essentially the idea is to move from the usual first-order data:

```haskell
data Person = Person { name :: Text, age :: Int }
```

to a [higher-kinded version](https://reasonablypolymorphic.com/blog/higher-kinded-data/) in which every field is wrapped in a type constructor:

```haskell
data Person f = Person { name :: f Text, age :: f Int }
```

Now we can indicate that every field is optional by using `Person Maybe`, or required by setting `Person Id`. As I said before, it also opens possibilities to describe things like "I want this field":

```haskell
data Wanted   a = Wanted | NotWanted deriving Functor
data Obtained a = HereYouAre a | DidNotAskForIt deriving Functor

queryGraphQL :: Person Wanted -> Person Obtained
```

Of course, I am more than happy to execute the plan.

serras:
As outlined in #39, I think it's worth investigating whether using HKD help us specify better what our code does. For example, we know that Protocol Buffers mandate that every field is optional: this new design allows us to be explicit about this fact.
In particular, `HasSchema` now has an additional parameter which is the "wrapper" type. You can set it to a particular type (like done in the `health-check` example, in which it's set to `Maybe`), or you can be parametric over it (as done in the examples in `core/rpc`).

Fixes #39

### DIFF
--- a/core/schema/src/Data/Functor/MaybeLike.hs
+++ b/core/schema/src/Data/Functor/MaybeLike.hs
@@ -1,0 +1,14 @@
+module Data.Functor.MaybeLike where
+
+import Data.Functor.Identity
+
+class MaybeLike f where
+  likeMaybe :: f a -> Maybe a
+
+instance MaybeLike Identity where
+  likeMaybe = Just . runIdentity
+instance MaybeLike Maybe where
+  likeMaybe = id
+instance MaybeLike (Either a) where
+  likeMaybe (Left  _) = Nothing
+  likeMaybe (Right y) = Just y
--- a/core/schema/src/Data/Functor/MaybeLike.hs
+++ b/core/schema/src/Data/Functor/MaybeLike.hs
@@ -12,3 +12,6 @@ instance MaybeLike Maybe where
 instance MaybeLike (Either a) where
   likeMaybe (Left  _) = Nothing
   likeMaybe (Right y) = Just y
+
+class FillDefaults f where
+  fillDefaults :: MaybeLike w => f w -> f Identity
--- a/core/schema/src/Data/Functor/MaybeLike.hs
+++ b/core/schema/src/Data/Functor/MaybeLike.hs
@@ -12,6 +12,3 @@ instance MaybeLike Maybe where
 instance MaybeLike (Either a) where
   likeMaybe (Left  _) = Nothing
   likeMaybe (Right y) = Just y
-
-class FillDefaults f where
-  fillDefaults :: MaybeLike w => f w -> f Identity
--- a/core/schema/src/Mu/Adapter/Json.hs
+++ b/core/schema/src/Mu/Adapter/Json.hs
@@ -17,7 +17,6 @@ import           Data.Aeson.Types
 import           Data.Functor.Contravariant
 import           Data.Functor.Identity
 import qualified Data.HashMap.Strict                 as HM
--- import           Data.SOP                            (NP (..), NS (..))
 import qualified Data.Text                           as T
 import qualified Data.Vector                         as V
 
--- a/core/schema/src/Mu/Adapter/Json.hs
+++ b/core/schema/src/Mu/Adapter/Json.hs
@@ -17,7 +17,7 @@ import           Data.Aeson.Types
 import           Data.Functor.Contravariant
 import           Data.Functor.Identity
 import qualified Data.HashMap.Strict                 as HM
-import           Data.SOP                            (NP (..), NS (..))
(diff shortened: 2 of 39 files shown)


core/schema/src/Data/Functor/MaybeLike.hs
```diff
@@ -0,0 +1,14 @@
+module Data.Functor.MaybeLike where
+
+import Data.Functor.Identity
+
+class MaybeLike f where
```
kutyel: Why do we need this new typeclass? 🤔

core/schema/src/Mu/Adapter/Json.hs
```diff
@@ -15,21 +15,22 @@ import           Control.Applicative                 ((<|>))
 import           Data.Aeson
 import           Data.Aeson.Types
 import           Data.Functor.Contravariant
+import           Data.Functor.Identity
 import qualified Data.HashMap.Strict                 as HM
-import           Data.SOP                            (NP (..), NS (..))
+-- import           Data.SOP                            (NP (..), NS (..))
```
kutyel: More commented code 🙃

core/schema/src/Mu/Schema/Conversion/SchemaToTypes.hs
```diff
@@ -45,71 +45,81 @@ typeDefToDecl :: Type -> Namer -> TypeDefB Type String String -> Q [Dec]
...
+       let hsi = generateHasSchemaInstance wTy schemaTy name complete (choiceMapping complete choices)
        return [d, hsi]
 -- Simple things
 typeDefToDecl _ _ (DSimple _)
   = fail "DSimple is not supported"
 
 deriveClauses :: [Q DerivClause]
 deriveClauses
-  = [ pure (DerivClause Nothing [ConT ''Eq, ConT ''Ord, ConT ''Show, ConT ''Generic]) ]
+  = [ pure (DerivClause Nothing [{- ConT ''Eq, ConT ''Ord, ConT ''Show, -} ConT ''Generic]) ]
```
kutyel: What is this magic? 💥

examples/todolist/src/Server.hs
```diff
@@ -39,49 +40,53 @@ reset i t = alwaysOk $ do
   atomically $ do
     writeTVar i 0
     writeTVar t []
-  pure $ MessageId 0 -- returns nothing
+  pure $ MessageId Nothing -- returns nothing
```
kutyel: Perfect, now all makes sense! Maybe we can remove the comment now 😉

core/schema/src/Data/Functor/MaybeLike.hs
serras: The idea is to give a uniform interface which could work for both `Identity` and `Maybe`. In theory everything could be done using monadic combinators, but by using `likeMaybe` as a view pattern, we can give a nice way to pattern match directly on everything.
Essentially, this tells you: if you can write your server assuming that field may be missing (because they are wrapped in a `MaybeLike` functor), you can make your server work on gRPC, and in the future, as Avro thingies.

examples/todolist/src/Server.hs
serras: Yay!

core/schema/src/Mu/Schema/Conversion/SchemaToTypes.hs
serras: This is something I need to fix still. Since now turning schemas to types creates a type with a `f (...)` inside, we need to use standalone deriving in order to add an `Applicative f` constraint. I need to figure out how to do this.

core/schema/src/Data/Functor/MaybeLike.hs
cb372: This caught my eye as well, but your explanation makes sense.

core/schema/src/Mu/Schema/Conversion/SchemaToTypes.hs
serras: It turns out that making these instances work correctly would be ton of work, for a module whose use is not recommended. So I've just disabled the generation for now.

PR #47, opened 2019-12-17
Extract `mu-persistent` to it's own package! ✂️

### LINKED ISSUE #45: Separate generic Persistent code into its own library
Currently, the `with-persistent` example contains some code which could be of interest of other users of Mu + Persistent. The goal would be to create a new library with that code.

kutyel:
Fixes #45

### DIFF
--- a/adapter/persistent/mu-persistent.cabal
+++ b/adapter/persistent/mu-persistent.cabal
@@ -5,7 +5,7 @@ version:             0.1.0.0
 homepage:            https://github.com/higherkindness/mu-haskell/persistent#readme
 license:             Apache-2.0
 license-file:        LICENSE
-author:              Flavio Corpa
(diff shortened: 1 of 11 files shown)


adapter/persistent/mu-persistent.cabal
```diff
@@ -0,0 +1,23 @@
...
+extra-source-files:  README.md
+
+library
+  exposed-modules:     Mu.Adapter.Persistent
+  hs-source-dirs:      src
+  default-language:    Haskell2010
+  build-depends:       base >= 4.7 && < 5
+                     , mu-schema
+                     , persistent
+                     , persistent-sqlite
```
serras: We should not require a specific database library here. I think this can be removed without any problem (other than adjusting imports).

PR #48, opened 2019-12-17
Replace Travis with Github Actions and Nix ❄️

### LINKED ISSUE #22: Replace Travis CI with Github Actions (Stack)
Since [some](https://github.com/arrow-kt/arrow/pull/1694/files) projects of 47 are moving to Github Actions, I think it would be a great idea to do the same!

Besides, I think it'd a great chance to **create our own Github action** for [Stack](https://docs.haskellstack.org/en/stable/README/) (since Github's one uses Cabal 3.0) and the rest of the Haskell community would benefit from it! 🎉 (maybe @gimenete can provide some guidance on how to do this 😉)

kutyel:
Huge thanks to @domenkozar for the help! 😉 🙌 

Fixes #22


.github/workflows/haskell.yml
```diff
@@ -0,0 +1,12 @@
...
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: cachix/install-nix-action@v6
+    - uses: cachix/cachix-action@v2
+      with:
+        name: kutyel
```
pepegar: Does this mean we're publishing to cachix under `kutyel`?

serras:
It seems, looking at the logs, that Nix uses a different root directory when compiling:

```
2019-12-17T16:00:50.5995315Z src/Mu/Adapter/ProtoBuf/Example.hs:1:1: error:
2019-12-17T16:00:50.5999118Z     Exception when trying to run compile-time code:
2019-12-17T16:00:50.6001605Z       test/protobuf/example.proto: openFile: does not exist (No such file or directory)
2019-12-17T16:00:50.6003989Z     Code: protobuf
2019-12-17T16:00:50.6016555Z             "ExampleProtoBufSchema" "test/protobuf/example.proto"
2019-12-17T16:00:50.6016676Z   |
2019-12-17T16:00:50.6017105Z 1 | {-# language DataKinds       #-}
```

default.nix
```diff
@@ -0,0 +1,21 @@
...
+  haskellnix = import (builtins.fetchTarball https://github.com/input-output-hk/haskell.nix/archive/master.tar.gz);
+  overlay = _: pkgs:
+    let
+      hnPkgs = pkgs.haskell-nix.stackProject {
+        src = ./.;
+        #inherit stackYaml;
+        modules = [];
+      };
+    in {
+      mu-grpc-server = hnPkgs.mu-grpc-server.components.all;
```
serras: Why `mu-grpc-server` explicitly?

.github/workflows/haskell.yml
kutyel: I'm afraid yes, should I create a 47deg cachix account @juanpedromoreno? 😉

domenkozar:
The downside of haskell.nix is that you have to compile GHC and everything up the to libraries first. It's a one time build you have to pay in order to get the benefit of not updating Nix when project goes on.

There's a way to use haskell.nix cachix to speed that up, but you'd need to pin the correct nixpkgs in that case (which may be ok with you?).

default.nix
kutyel: I have the same doubt 😂 @domenkozar? 😉

kutyel:
> It seems, looking at the logs, that Nix uses a different root directory when compiling:
> 
> ```
> 2019-12-17T16:00:50.5995315Z src/Mu/Adapter/ProtoBuf/Example.hs:1:1: error:
> 2019-12-17T16:00:50.5999118Z     Exception when trying to run compile-time code:
> 2019-12-17T16:00:50.6001605Z       test/protobuf/example.proto: openFile: does not exist (No such file or directory)
> 2019-12-17T16:00:50.6003989Z     Code: protobuf
> 2019-12-17T16:00:50.6016555Z             "ExampleProtoBufSchema" "test/protobuf/example.proto"
> 2019-12-17T16:00:50.6016676Z   |
> 2019-12-17T16:00:50.6017105Z 1 | {-# language DataKinds       #-}
> ```

I just added some changes that should fix this, thanks again @domenkozar 🎉 

For reference, apparently cabal does not like external files, so you need to add them to your `data-files `entry on your cabal, I followed this [post](http://neilmitchell.blogspot.com/2008/02/adding-data-files-using-cabal.html) by Neil Mitchell 👏

kutyel:
@juanpedromoreno it seems like the 47deg.cachix.org account is not correctly configured:
```json
responseBody = "{\"error\":\"No public key set yet. Run 'cachix generate-keypair' first.\"}"})
```
Can you please take a look? 😉

kutyel:
Ok, it is fixed now 🎉  It should work! 🤞

PR #49, opened 2019-12-19
Make ghcide happy

serras:
This PR adds a `hie.yaml` file to each package, which instructs `ghcide` about which packages to load in each case. In addition, it uses the new `__GHCIDE__` macro to use different paths during IDE operation and type checking.

### DIFF
--- a/adapter/avro/src/Mu/Adapter/Avro/Example.hs
+++ b/adapter/avro/src/Mu/Adapter/Avro/Example.hs
@@ -1,3 +1,4 @@
+{-# language CPP         #-}
 {-# language DataKinds   #-}
 {-# language QuasiQuotes #-}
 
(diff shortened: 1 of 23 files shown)


adapter/avro/src/Mu/Adapter/Avro/Example.hs
```diff
@@ -35,4 +36,8 @@ type Example = [avro|
 }
 |]
 
+#if __GHCIDE__
```
kutyel: Can we somehow make all paths to files work for ghcide/cabal/stack/nix, etc... by tweaking somehow the `*.cabal` file? 🤔 Maybe that way we can skip using `CPP` all over the examples.

serras: I don't know how to do it without using even more CPP magic :(

PR #50, opened 2019-12-19
Document integration with Persistent 📚

### LINKED ISSUE #38: Document how to integrate with Persistent
Once #30 is finished, create a tutorial with the steps required to use Persistent with mu-haskell

kutyel:
Fixes #38

### DIFF
--- a/docs/db.md
+++ b/docs/db.md
@@ -23,7 +23,7 @@ service PersistentService {
 }
 ```
 
-Maybe this example looks a bit contrived but bear with me, it covers a common use case when working with protobuf: that one of the messages has another message as its identifying key. Easier examples would not suffer from this and should be easier to implement.
+Maybe this example looks a bit contrived but bear with me, it covers a common use case when working with protobuf: that one of the messages has another message as its identifying key.
 
 ## Definning our Schema
 
@@ -33,7 +33,6 @@ You are going to need to enable the following extensions:
 {-# language DataKinds                  #-}
 {-# language DeriveGeneric              #-}
 {-# language DerivingVia                #-}
-{-# language DuplicateRecordFields      #-}
 {-# language FlexibleContexts           #-}
 {-# language FlexibleInstances          #-}
 {-# language GADTs                      #-}
@@ -74,7 +73,7 @@ instance FromSchema Maybe PersistentSchema "Person" MPerson
 
 Remember that all the magic starts with that first `grpc` line! ✨
 
-You might have noticed that this time, we are not using `DeriveAnyClass`, so we need to write the instances for `ToSchema` and `FromSchema` on a separate line from our deriving clause, and let GHC fill them for us.
+You might have noticed that this time, we are not using `DeriveAnyClass`, so we need to write the instances for `ToSchema` and `FromSchema` on a separate line from our deriving clause, and let GHC fill them for us. This decision was made due to a current [bug in Persistent](https://github.com/yesodweb/persistent/issues/578), but hopefully it will be fixed in future versions. 🙂
 
 ## Integration with `persistent`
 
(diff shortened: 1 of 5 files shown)


docs/db.md
```diff
@@ -1,4 +1,200 @@
...
+message Person { PersonRequest pid = 1; string name = 2; int32 age = 3; }
+
+service PersistentService {
+  rpc getPerson (PersonRequest) returns (Person);
+  rpc newPerson (Person) returns (PersonRequest);
+  rpc allPeople (google.protobuf.Empty) returns (stream Person);
+}
+```
+
+Maybe this example looks a bit contrived but bear with me, it covers a common use case when working with protobuf: that one of the messages has another message as its identifying key. Easier examples would not suffer from this and should be easier to implement.
```
serras: It's not very clear what "this" refers to in "easier examples would not suffer from this".

docs/db.md
```diff
@@ -1,4 +1,200 @@
...
+
+## Definning our Schema
+
+You are going to need to enable the following extensions:
+
+```haskell
+{-# language DataKinds                  #-}
+{-# language DeriveGeneric              #-}
+{-# language DerivingVia                #-}
+{-# language DuplicateRecordFields      #-}
```
serras: I'm not sure we need that one.

docs/db.md
```diff
@@ -1,4 +1,200 @@
...
+  , age  :: Maybe Int32 }
+  deriving (Eq, Ord, Show, Generic)
+
+instance ToSchema   Maybe PersistentSchema "Person" MPerson
+instance FromSchema Maybe PersistentSchema "Person" MPerson
+```
+
+Remember that all the magic starts with that first `grpc` line! ✨
+
+You might have noticed that this time, we are not using `DeriveAnyClass`, so we need to write the instances for `ToSchema` and `FromSchema` on a separate line from our deriving clause, and let GHC fill them for us.
```
serras: Maybe say that this choice is due to the [following Persistent bug](https://github.com/yesodweb/persistent/issues/578)?

docs/db.md
```diff
@@ -1,4 +1,200 @@
...
+  runStderrLoggingT $
+    withSqliteConn @(LoggingT IO) "example.db" $ \conn ->
+      liftIO $ runGRpcApp 8080 (server conn)
+```
+
+We have decided in this example to use `LoggintT` from `monad-logger` and `runStderrLoggingT` to get some basic database logs to the console for free, but this is not a must!
+
+## This actually does not work
+
+Maybe you might have noticed that this example is not going to work yet, until we tell `persistent` what is our migration. We need a small tweak in our `Schema.hs`:
```
serras: This is unless you have created the `example.db` file yourself. Migrations are not actually required by Persistent, are just a simple way to get a Sqlite database up and running.

docs/db.md
```diff
@@ -1,4 +1,200 @@
...
+
+As you can see, all the services need to be passed the `SqlBackend` connection as an argument.
+
+Two interesting things we want to highlight here: we have provided a small helper called `runDb`, it's implementation is quite simple and it exist due to **developer ergonomics**. We are basically saving you from writing lots of `liftIO $ flip runSqlPersistM`. 😉
+
+The second one will be discussed in the next section.
+
+## On streams and `Conuit`
+
+Since we are going to work with streams, it is wonderful that `persistent` also provides methods to work with `Conduit` like, for example, `selectSource`. But the Monad in which `persistent` operates returns `ConduitM () (Entity Person) m ()`, and we know that we want instead: `ConuitT (Entity Person) Void ServerErrorIO ()`. 🤔
```
serras: Typo: `Conuit` instead of `Conduit`.

docs/db.md
```diff
@@ -1,4 +1,200 @@
...
+
+Well, have no fear my friend because we created yet another utility called `liftServerConduit`. Born specifically to address this problem, it's type signature is:
```
serras: What is this type signature telling us. I would add something below in the lines of:

> That is, we can turn any of the conduits given as input, which work on the `ServerErrorIO` monad from `mu-rpc`, into a conduit working on other `IO`-like monad. This is the case, in particular, of the monad in which Persistent runs.

PR #51, opened 2019-12-19
Update to v3 of cachix-action ❄️

### LINKED ISSUE #57: Please don't use `master` archive of haskell.nix, use a specific reference.
Great to see this using haskell.nix! Looking thorugh the code, I've seen this line:

https://github.com/higherkindness/mu-haskell/blob/1b6f511acfa93d40f930c2361fec99352c7f1696/default.nix#L1

I'd like to caution against using the `master` archive, and rather put in a git ref instead. Otherwise it's impossible to tell against which version it was built, and any form of issue bisection will be near impossible.

Using https://github.com/input-output-hk/haskell.nix/archive/83966f3.tar.gz would be much better.

kutyel:
A follow up on #48 

Hopefully this should go faster 🤞 

cc @domenkozar 😉 

---

Fixes #57


domenkozar:
Try only amending commit to see how much faster is a noop.

kutyel:
@domenkozar something wrong is going on 😕 . Each time it compiles the entire GHC 😂

domenkozar:
Yes, there was a bug with pushing. Blame JavaScript :) Can you amend to try again, that will start pushing everything to Cachix, so second amend should be then quick.

domenkozar:
2m! Now you'll need to manually list all packages you want to build until https://github.com/input-output-hk/haskell.nix/issues/377 is implemented.

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,12 @@
+name: Haskell CI
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: cachix/install-nix-action@v6
+    - uses: cachix/cachix-action@releases/v3
```
domenkozar:
```suggestion
    - uses: cachix/cachix-action@v3
```

default.nix
```diff
@@ -0,0 +1,24 @@
+{ pkgs ? import <nixpkgs> (import (builtins.fetchTarball https://github.com/input-output-hk/haskell.nix/archive/master.tar.gz))
```
domenkozar: Ideally you'll want to pin nixpkgs as well. Otherwise you'll need to keep rebuilding GHC, etc everytime channel advanced.

default.nix
```diff
@@ -0,0 +1,24 @@
...
+, system ? builtins.currentSystem
```
domenkozar:
```suggestion
```

kutyel:
We did it people, builds of <5min! 👏👏 👏  Thanks again @domenkozar for everything, Cachix is great! 🎉

domenkozar:
My last recommendation would be to pin `nixpkgs`:

```
{ nixpkgs ? (fetchTarball https://github.com/NixOS/nixpkgs/archive/b1844ef5816b0af8bc2f6215054279ea35e29b77.tar.gz)
, pkgs ? import nixpkgs (import (builtins.fetchTarball https://github.com/input-output-hk/haskell.nix/archive/master.tar.gz))
}:
```

This way CI is completely reproducible.

PR #61, opened 2020-01-02
Document gRPC client development! 📚

### LINKED ISSUE #58: Document gRPC client development
There are currently two different ways of building clients:

1. Using records
2. Using `TypeApplications`

Both should be documented [here](https://github.com/higherkindness/mu-haskell/blob/master/docs/docs/grpc.md).

kutyel:
Fixes #58

### DIFF
--- a/docs/docs/grpc.md
+++ b/docs/docs/grpc.md
@@ -72,11 +72,13 @@ main = do
    let config = grpcClientConfigSimple "127.0.0.1" 1234 False
 -  Right client <- setupGrpcClient' config
 +  Right grpcClient <- setupGrpcClient' config
-+  let client = buildService @Service @"" grpcClient
++  let client = buildService @Service "/grpc" grpcClient
    args <- getArgs
 ```
 
-Instead of building our client directly, we can use `TypeApplications` to call `buildService` to create the actual gRPC client. (You can also use `ScopedTypeVariables` and something like `(Proxy :: Proxy Service)` but it's a bit more verbose).
+Instead of building our client directly, we need to call `buildService` (and enable `TypeApplications`) to create the actual gRPC client. If you don't like `TypeApplications`, it is also possible to use a combination of `ScopedTypeVariables` and something like `(Proxy :: Proxy Service)` to achieve the same result, but it's a bit more verbose and we _encourge you<sup>1</sup>_ to use `TypeApplications`  instead. 😉
+
+That string (or `ByteString`) as a second argument to `buildService` corresponds to the route of the service.
 
 After that, let's have a look at an example implementation of the three service calls:
 
@@ -147,5 +149,7 @@ watching client = do
 Here though, while mapping `print` to the `Conduit`, we needed to add a type annotation because the type was ambiguous... I think it's a small price to pay in exchange for the terseness. 🤑
 
 ---
+
+<sup>1</sup> To read more on `TypeApplications`, you can check [this](https://www.reddit.com/r/haskell/comments/6ufnmr/scrap_your_proxy_arguments_with_typeapplications/), [that](https://blog.sumtypeofway.com/posts/fluent-polymorphism-type-applications.html) and [this](https://kseo.github.io/posts/2017-01-08-visible-type-application-ghc8.html).
 
 To see a **working example** you can check all the code at the [example with persistent](https://github.com/higherkindness/mu-haskell/tree/master/examples/with-persistent).
--- a/docs/docs/grpc.md
+++ b/docs/docs/grpc.md
@@ -18,7 +18,7 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
 
 ## Building a client
 
-Right now there are two options for building clients: using records or with `XTypeApplications`. To give a proper introduction to both options let's consider in detail an example client for the following services:
+Right now there are two options for building clients: using records or with `TypeApplications`. To give a proper introduction to both options let's consider in detail an example client for the following services:
 
 ```protobuf
 service Service {
@@ -49,9 +49,9 @@ Where `watch`, `get` and `add` are the only valid 3 commands that our CLI is goi
 
 ### Using records
 
-This option is a bit more verbose but it's also more explicit with the types and _"less magic"_ than the one with `TypeApplications`.
+This option is a bit more verbose but it's also more explicit with the types and _"a bit more magic"_ than the one with `TypeApplications` (due to the use of Generics).
 
-We need to define a new record type (hence the name) that declares the services our client is going to consume:
+We need to define a new record type (hence the name) that declares the services our client is going to consume. Remember that the names of the record fields **must match** exactly the methods in the service:
 
 ```haskell
 import GHC.Generics (Generic)
@@ -76,9 +76,9 @@ main = do
    args <- getArgs
 ```
 
-Instead of building our client directly, we need to use `TypeApplications` (yes, we can't escape from `TypeApplications` either way 🤷🏼‍♂️) in a call to `buildService` to create the actual gRPC client.
+Instead of building our client directly, we can use `TypeApplications` to call `buildService` to create the actual gRPC client. (You can also use `ScopedTypeVariables` and something like `(Proxy :: Proxy Service)` but it's a bit more verbose).
 
-After that, the implementation of the three service calls becomes quite trivial:
+After that, let's have a look at an example implementation of the three service calls:
 
 ```haskell
 import Text.Read (readMaybe)
@@ -135,7 +135,7 @@ add client nm ag = do
   putStrLn $ "ADD: was creating successful? " ++ show response
 ```
 
-We are being a bit more explicit with the types (for example, `response :: GRpcReply MPersonRequest`) to help a bit the `show` function with the ambiguity.
+We are being a bit more explicit with the types here (for example, `response :: GRpcReply MPersonRequest`) to help a bit the `show` function because GHC is not able to infer the type on its own.
 
 ```haskell
 watching :: GrpcClient -> IO ()
--- a/docs/docs/grpc.md
+++ b/docs/docs/grpc.md
@@ -91,7 +91,7 @@ get client idPerson = do
   putStrLn $ "GET: response was: " ++ show res
 ```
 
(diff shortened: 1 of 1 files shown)


docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+get :: Call -> String -> IO ()
+get client idPerson = do
+  let req = MPersonRequest $ readMaybe idPerson
+  putStrLn $ "GET: is there some person with id: " ++ idPerson ++ "?"
+  res <- getPerson client req
+  putStrLn $ "GET: response was: " ++ show res
+```
 
-### Using `TypeApplications`
+Notice the use of `readMaybe` to cast the strings to the appropiate type in a safe manner! 👆🏼
```
pepegar:
```suggestion
Notice the use of `readMaybe` to convert the strings to the appropiate type in a safe manner! 👆🏼
```

Just to avoid people thinking we're `unsafeCoerce` ing here.

kutyel: Thanks pepe! ❤️

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+
+Note that we had to derive `Generic`. We also need to tweak a little bit our `main` function:
+
+```diff
+main :: IO ()
+main = do
+   let config = grpcClientConfigSimple "127.0.0.1" 1234 False
+-  Right client <- setupGrpcClient' config
++  Right grpcClient <- setupGrpcClient' config
++  let client = buildService @Service @"" grpcClient
```
cb372: I'm a bit confused by `@""`. `""` is a singleton type?

Is there some other documentation where I can check the signature of `buildService`? If so, a link to that would be good.

kutyel: I agree, that bit is quite confusing, [here's the type signature](https://github.com/higherkindness/mu-haskell/blob/master/grpc/client/src/Mu/GRpc/Client/Record.hs#L46-L50) of `buildService` but you are not gonna like it 😂. I'll wait for Alejandro's feedback when he comes back to see what else to add/clarify on the docs 😉

cb372: Art.

serras: The `@""` is a type-level string which defines the prefix of the service, something akin to the route where the service is located. I have been doubting for a while whether a type-level string is the best choice, or we should use a normal string (well, actually a `ByteString`) here instead.

We still have time to change the API, and we can even provide both versions.

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
 
 ## Building a client
 
--- TODO:
+Right now there are two options for building clients: using records or with `XTypeApplications`. To give a proper introduction to both options let's consider in detail an example client for the following services:
```
serras: Usually, one writes either `TypeApplications` (because that's the parameter to `LANGUAGE` pragmas) or `-XTypeApplications` (because that's the one you enable it in ghci).

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+    ["add", nm, ag] -> add client nm ag
+    _               -> putStrLn "unknown command"
+```
+
+Where `watch`, `get` and `add` are the only valid 3 commands that our CLI is going to accept and call each respective service.
 
 ### Using records
 
--- TODO:
+This option is a bit more verbose but it's also more explicit with the types and _"less magic"_ than the one with `TypeApplications`.
```
serras: I would say this is more magic, due to the use of generics xD

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+
+We need to define a new record type (hence the name) that declares the services our client is going to consume:
```
serras: It's important to emphasize that the names of the fields **must** coincide with the methods in the service.

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+main :: IO ()
+main = do
+   let config = grpcClientConfigSimple "127.0.0.1" 1234 False
+-  Right client <- setupGrpcClient' config
++  Right grpcClient <- setupGrpcClient' config
++  let client = buildService @Service @"" grpcClient
+   args <- getArgs
+```
+
+Instead of building our client directly, we need to use `TypeApplications` (yes, we can't escape from `TypeApplications` either way 🤷🏼‍♂️) in a call to `buildService` to create the actual gRPC client.
```
serras: That's not completely true, you can call the version using `Proxy`. But then you need to enable `ScopedTypeVariables` and use `(Proxy :: Proxy Service)`, which is even more verbose and harder to understand.

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+
+After that, the implementation of the three service calls becomes quite trivial:
```
serras: Usual comment applies here: try to avoid words like "simply" or "trivial", because the reader may not feel that way.

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+add :: GrpcClient -> String -> String -> IO ()
+add client nm ag = do
+  let p = MPerson Nothing (Just $ T.pack nm) (readMaybe ag)
+  putStrLn $ "ADD: creating new person " ++ nm ++ " with age " ++ ag
+  response :: GRpcReply MPersonRequest
+    <- gRpcCall @Service @"newPerson" client p
+  putStrLn $ "ADD: was creating successful? " ++ show response
+```
+
+We are being a bit more explicit with the types (for example, `response :: GRpcReply MPersonRequest`) to help a bit the `show` function with the ambiguity.
```
serras: I would rephrase this as "because GHC is not able to infer the type by its own".

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
 
 ## Building a client
 
--- TODO:
+Right now there are two options for building clients: using records or with `TypeApplications`. To give a proper introduction to both options let's consider in detail an example client for the following services:
```
diesalbla:
```suggestion
There are two ways to build a gRPC client: 1) using a generic Haskell records, or 2) using the `TypeApplications` GHC extension. To show how to use each of them, and how they differ, let us consider an example client for the following services:
```

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+
+```protobuf
+service Service {
+  rpc getPerson (PersonRequest) returns (Person);
+  rpc newPerson (Person) returns (PersonRequest);
+  rpc allPeople (google.protobuf.Empty) returns (stream Person);
+}
+```
+
+Regardless of the approach we decide to use, we can construct a basic CLI for the client this way:
```
diesalbla:
```suggestion
In both approaches, we can build a basic CLI program around our client, using the following `main` function: 
```

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+```
+
+Where `watch`, `get` and `add` are the only valid 3 commands that our CLI is going to accept and call each respective service.
 
 ### Using records
 
--- TODO:
+This option is a bit more verbose but it's also more explicit with the types and _"a bit more magic"_ than the one with `TypeApplications` (due to the use of Generics).
+
+We need to define a new record type (hence the name) that declares the services our client is going to consume. Remember that the names of the record fields **must match** exactly the methods in the service:
```
diesalbla:
```suggestion
In this approach, we define a record type to represent the service call. Each field in this record corresponds to an endpoint (a method in the remote service). The name of each field **must match** the name of the corresponding endpoint:
```

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+main :: IO ()
+main = do
+   let config = grpcClientConfigSimple "127.0.0.1" 1234 False
+-  Right client <- setupGrpcClient' config
++  Right grpcClient <- setupGrpcClient' config
++  let client = buildService @Service @"" grpcClient
+   args <- getArgs
+```
+
+Instead of building our client directly, we can use `TypeApplications` to call `buildService` to create the actual gRPC client. (You can also use `ScopedTypeVariables` and something like `(Proxy :: Proxy Service)` but it's a bit more verbose).
```
diesalbla: I am confused here... this subsection is about the generic record approach; so is the `TypeApplications` also used with the generic records? Or should this go to the next section?

docs/docs/grpc.md
```diff
@@ -18,12 +18,134 @@ main = runGRpcApp 8080 "helloworld" quickstartServer
...
+  putStrLn $ "GET: is there some person with id: " ++ idPerson ++ "?"
+  response :: GRpcReply MPerson
+    <- gRpcCall @Service @"getPerson" client req
+  putStrLn $ "GET: response was: " ++ show response
+```
+
+Notice that the type signatures of our functions needed to change to receive the `GrpcClient` as an argument, instead of our custom record type.
+
+```haskell
+add :: GrpcClient -> String -> String -> IO ()
```
diesalbla: Could we factor out the implementation of `add` above, next to the `main` function, and make it parametric on a `Person -> IO String` function? Same for the `get` function.

docs/docs/grpc.md
kutyel: Yes, if you want the less verbose option, you need to use `TypeApplications` to generate the client with `buildService`, even when you choose the record approach, that's why the original wording reflects that surprising fact.

PR #67, opened 2020-01-09
Add auto-deploy of microsite to GH Pages 🐙

kutyel:
Related to https://github.com/47deg/marlow/issues/406

I can't generate haddock docs yet because CI does not know about stack, maybe we chould find a way to generate the docs with Nix 🤔 (although I think the current `cachix-action` is already generating the haddock docs somewhere).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
     - master
-    - site-deploy-gh
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,5 +25,5 @@ jobs:
       uses: peaceiris/actions-gh-pages@v2
       env:
         ACTIONS_DEPLOY_KEY: ${{ secrets.ACTIONS_DEPLOY_KEY }}
-        PUBLISH_BRANCH: master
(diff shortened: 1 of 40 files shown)


calvellido:
Maybe this could be added to get Haskell in the mix:

https://github.com/actions/setup-haskell

kutyel:
> Maybe this could be added to get Haskell in the mix:
> 
> https://github.com/actions/setup-haskell

I'm afraid that is the default GitHub action for Haskell, which uses Cabal 3.0, and right now our Haddock generations relies entirely on Stack 😕 

I don't think it would be ideal to mix build tools in the same CI 😂

kutyel:
I see that it does not understand the `bundle` command... @calvellido should we try https://github.com/BryanSchuetz/jekyll-deploy-gh-pages instead? 🤔

rachelcarmena:
Hi mates! Some comments:

- To solve the problem, try to add a previous step to install that command:
```
sudo apt-get update
sudo apt-get install ruby-dev
sudo gem install bundler --force
sudo gem update --system
```
- Try to separate build and deploy. 
- Build on:
```
on: pull_request
```
- Deploy (it will include build as well) on:
```
on:
  push:
    branches:
    - master
```
- `PUBLISH_BRANCH: gh-pages` can be `master`
- An idea: create a repository on your personal account to make tests with custom actions very quickly.

PR #68, opened 2020-01-09
Update docs about gRPC clients

serras:


docs/docs/grpc.md
```diff
@@ -72,13 +72,11 @@ main = do
    let config = grpcClientConfigSimple "127.0.0.1" 1234 False
 -  Right client <- setupGrpcClient' config
 +  Right grpcClient <- setupGrpcClient' config
-+  let client = buildService @Service "/grpc" grpcClient
++  let client = buildService @Service @"call_" grpcClient
    args <- getArgs
 ```
 
-Instead of building our client directly, we need to call `buildService` (and enable `TypeApplications`) to create the actual gRPC client. If you don't like `TypeApplications`, it is also possible to use a combination of `ScopedTypeVariables` and something like `(Proxy :: Proxy Service)` to achieve the same result, but it's a bit more verbose and we _encourge you<sup>1</sup>_ to use `TypeApplications`  instead. 😉
-
```
kutyel: Since you removed the <sup>1</sup> from the docs, rephrase the footnote to recommend `TypeApplications` in the kindest way possible and also remove the number 1 from there too please 😉

serras: Done!

PR #97, opened 2020-01-27
Fix deploy: last tag and content

rachelcarmena:


rachelcarmena:
@calvellido , please, take a look at it because I removed unused files.

.github/workflows/deploy.yml
```diff
@@ -20,15 +20,17 @@ jobs:
...
       run: |
-        LAST_TAG=$(git describe)
+        LAST_TAG=$(git describe --abbrev=0)
         git checkout $LAST_TAG
         bash ./generate-haddock-docs.sh docs
         BUNDLE_GEMFILE=./docs/Gemfile bundle exec jekyll build -b /mu-haskell -s docs -d gen-docs
     - name: Build microsite from master
       run: |
         git checkout master
+        sed -i "s/haddock\\//wip\\/haddock\\//g" docs/_includes/_doc.html
```
calvellido: I don't think that these two `sed` are necessary. When the `relative_url` filter is set for a string/URL, they get built considering the base URL set through the `-b` param, otherwise we'd need to hardfix all routes in the site.

rachelcarmena: Good idea!! :+1:

rachelcarmena: Fixed @calvellido!

PR #98, opened 2020-01-27
Fix deploy: path

rachelcarmena:


.github/workflows/deploy.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       run: |
         git checkout master
         bash ./generate-haddock-docs.sh docs/wip
```
calvellido:
```suggestion
        bash ./generate-haddock-docs.sh docs
```

Like this

rachelcarmena: Good idea!! :+1: However, now I wonder if there will be conflicts with the previous content of `docs`. What do you think about it?

rachelcarmena: NVM, `docs/haddock` is removed as a first step in the script :+1: Thanks!

rachelcarmena: Fixed! Thanks @calvellido !!

PR #100, opened 2020-01-29
You can use `GITHUB_TOKEN` already provided by Github Actions

alejandrohdezma:
Reference: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -32,6 +32,6 @@ jobs:
     - name: Deploy microsite
       uses: peaceiris/actions-gh-pages@v2
       env:
-        ACTIONS_DEPLOY_KEY: ${{ secrets.ACTIONS_DEPLOY_KEY }}
+        ACTIONS_DEPLOY_KEY: ${{ secrets.GITHUB_TOKEN }}
         PUBLISH_BRANCH: gh-pages
         PUBLISH_DIR: ./gen-docs


serras:
Sorry for the question, since I'm far from familiar with GitHub Actions. Is this a better way to do it than what we were doing? Should we expect any changes in our pipeline with this change? Thanks for the PR :)

alejandrohdezma:
Sorry @serras I should've give more context, Github Actions already creates a `GITHUB_TOKEN` secret automatically for your workflows so you don't need to add one under `Settings` > `Secrets`

serras:
Unfortunately the deployment failed https://github.com/higherkindness/mu-haskell/runs/416664329?check_suite_focus=true. The error is:

```
Load key "/root/.ssh/id_rsa": invalid format
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

@alejandrohdezma could you have a look at this? In the meanwhile I'll revert the commit.

alejandrohdezma:
@serras I'll take a look at it

rachelcarmena:
Hi! According to the [documentation for that GitHub Action](https://github.com/peaceiris/actions-gh-pages) instead of:

```
ACTIONS_DEPLOY_KEY:
```

it should be:

 ```
GITHUB_TOKEN:
```

for using that token.

alejandrohdezma:
Hey @rachelcarmena! After diving into this issue I think I was mistaken in the first place. As said in the plugin documentation:

> Currently, GitHub Actions does not support to trigger a GitHub Pages build event using GITHUB_TOKEN on a public repository.

So I think we have to stick with the custom token for now :(

rachelcarmena:
Oh, right! I didn't read the `(1)` :woman_facepalming: sorry. Thanks!!

serras:
@alejandrohdezma @rachelcarmena thanks for clarifying the issue! I'll try to revert the PR, although it seems that the other part of our CI infrastructure (Cachix) is failing, so we might need to wait until @kutyel comes back from holidays.

PR #107, opened 2020-02-06
Support for decimals

### LINKED ISSUE #106: Support BigDecimal
Avro supports a [decimal](https://avro.apache.org/docs/1.9.0/spec.html#Decimal) type with arbitrary scale and precision. This is not supported by the `avro` library we are using, so we need to add it.

---

**Decimal**
The decimal logical type represents an arbitrary-precision signed decimal number of the form _unscaled × 10-scale._

A decimal logical type annotates Avro bytes or fixed types. The byte array must contain the two's-complement representation of the unscaled integer value in big-endian byte order. The scale is fixed, and is specified using an attribute.

The following attributes are supported:

scale, a JSON integer representing the scale (optional). If not specified the scale is 0.
precision, a JSON integer representing the (maximum) precision of decimals stored in this type (required).
For example, the following schema represents decimal numbers with a maximum precision of 4 and a scale of 2:
```
{
  "type": "bytes",
  "logicalType": "decimal",
  "precision": 4,
  "scale": 2
}
```
Precision must be a positive integer greater than zero. If the underlying type is a fixed, then the precision is limited by its size. An array of length n can store at most _floor(log_10(28 × n - 1 - 1))_ base-10 digits of precision.

Scale must be zero or a positive integer less than or equal to the precision.

For the purposes of schema resolution, two schemas that are decimal logical types match if their scales and precisions match.

serras:
Fixes #106


serras:
At the end most of the code is now in https://github.com/haskell-works/avro/pull/135, so I am opening a new PR

PR #110, opened 2020-02-17
Update to LTS 15

serras:


default.nix
```diff
@@ -1,5 +1,5 @@
-{ nixpkgs ? (fetchTarball https://github.com/NixOS/nixpkgs/archive/484ea7bbac5bf7f958b0bb314a3c130b6c328800.tar.gz)
-, pkgs ? import nixpkgs (import (builtins.fetchTarball https://github.com/input-output-hk/haskell.nix/archive/659d80f.tar.gz))
+{ nixpkgs ? (fetchTarball https://github.com/NixOS/nixpkgs/archive/497c0a43a5982081f8310e9c0b7fcf91d33c0001.tar.gz)
```
kutyel: Can we try using the short hash here? I've been wondering for some time if it will work 🤔

kutyel:
> This version of stackage.nix does not know about the Stackage resolver lts-15.0.
We'll have to wait 😄

kutyel:
We won't be able to merge this until https://github.com/input-output-hk/nix-tools/pull/78 gets merged!

stack.yaml
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
 resolver: lts-15.0
+=======
+resolver: lts-14.27
+>>>>>>> master
```
kutyel: I think git has messed with you a bit 😆

serras: That's why VS Code was asking me so many times "do you really want to commit?"

stack-nightly.yaml
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
+resolver: nightly-2020-02-17
+=======
 resolver: nightly-2020-02-28
+>>>>>>> master
```
kutyel: Here as well 😝

stack-14.yaml
```diff
@@ -0,0 +1,50 @@
...
+extra-deps:
+- http2-client-0.9.0.0
+- http2-grpc-types-0.5.0.0
+- proto3-wire-1.1.0
+- http2-grpc-proto3-wire-0.1.0.0
+- warp-grpc-0.3.0.0
+- http2-client-grpc-0.8.0.0
+# - avro-0.4.6.0
+- git: https://github.com/haskell-works/avro.git
+  commit: 93bc12fdb807e9567ef113ed71c7b3487516a371
```
kutyel:
```suggestion
- avro-0.4.7.0
```

stack-14.yaml
```diff
@@ -0,0 +1,50 @@
...
+- language-protobuf-1.0.1
+# - language-avro-0.1.0.0
+- git: https://github.com/higherkindness/avro-parser-haskell.git
+  commit: c8f7497766d0d1be8579fb03b302340adafeaf49
```
kutyel:
```suggestion
- language-avro-0.1.2.0
```

kutyel:
I think this is not yet going to work since there's no derivation for the `lts-15` in https://github.com/input-output-hk/stackage.nix... 🤷‍♂

serras:
It seems that LTS 15 are finally in the `stackage.nix` repo. Let's wait a couple of hours until GHC is built, but it seems good to do this time!

kutyel:
Maybe we can close this in favor of https://github.com/higherkindness/mu-haskell/pull/155, although... I'd still like to give Nix an opportunity when they fix `stakage.nix` for **lts-15**!

serras:
To be honest, I would prefer deprecating Nix if Stack works OK, since that way our dev environment is the same as the CI env.

serras:
Unfortunately we have decided to use Stack instead of Nix for CI, since we couldn't get it to work with LTS-15 :(

PR #119, opened 2020-03-02
Fix logic for getting last tag in GH Actions

serras:


serras:
Summoning @rachelcarmena because she might know a bit better how to solve this. My fix seems to work, at least locally, but I feel like I'm playing with dark magic here.

rachelcarmena:
> Summoning @rachelcarmena because she might know a bit better how to solve this. My fix seems to work, at least locally, but I feel like I'm playing with dark magic here.

Hi! Thanks for your trust @serras !! :heart: 

However, I don't recommend this change :sweat_smile: Why? 

It's necessary now because some new tags are not **annotated** but **lightweight** (created without `-a`).

> A lightweight tag is very much like a branch that doesn’t change — it’s just a pointer to a specific commit.

> Annotated tags, however, are stored as full objects in the Git database. They’re checksummed; contain the tagger name, email, and date; have a tagging message; and can be signed and verified with GNU Privacy Guard (GPG). It’s generally recommended that you create annotated tags so you can have all this information; but if you want a temporary tag or for some reason don’t want to keep the other information, lightweight tags are available too.

I've seen your tags:

```
rachel@nobrain:~/lab/mu-haskell$ git show-ref -d --tags 
fd11ed68f03dc1548c73a3ab34d3c5025a17e174 refs/tags/0.1.2
b42463ae1fac6dbccb61afb15dc9ec4a92c53346 refs/tags/v0.1
c24a55d83c7786ef040f188cb68f8776fad6a41a refs/tags/v0.1^{}
ef1430f6db39c5d889c35e24d63859348733752d refs/tags/v0.1.1
733b24b7ea4367df603626172075eb4abbdb04f3 refs/tags/v0.1.1^{}
e7592933e2331123abe33b4abcb8194996ce4e63 refs/tags/v0.2
d6815e66d4486d975be2b7d1fcece866b2aaeaf4 refs/tags/v0.2.0.1
```

Just 2 tags are annotated (showed twice with `^{}`).

My humble recommendation:
* Have a clear tag format. There are several mixed formats now:
```
rachel@nobrain:~/lab/mu-haskell$ git tag -l
0.1.2
v0.1
v0.1.1
v0.2
v0.2.0.1
```
* Convert lightweight tags to annotated tags.
```
# To keep the same tag name:
git tag -a -f <tagname> <tagname>
git push -f origin <tagname>

# To use a different tag name:
git tag -a <new-tagname> <existing-lightweight-tag>
git tag -d <existing-lightweight-tag>
git push origin <new-tagname>
git push origin -d <existing-lightweight-tag>
```
* Undo this change :sweat_smile:

rachelcarmena:
And create annotated tags (with `-a`) :wink:

rachelcarmena:
When adding ` --tags` all the tags are considered, not only annotated but also lightweight.

serras:
Wow! Sometimes Git has some subtleties I wasn't really expecting, to be honest.

The crux of the problem is that we are using the *Releases* functionality in GH to create a tag along with a release, and it seems that it's using lightweight tags. The fact that we can have a list of releases directly on the repo is something we don't want to lose 😿

rachelcarmena:
Oh! I didn't know that GitHub created lightweight tags... :open_mouth: I usually create the tag and then I use _Releases_ functionality:

Screenshot from 2020-03-02 11-32-31

PR #130, opened 2020-03-09
Remove higher-kinded data

serras:
In the previous version we used [higher-kinded data](https://reasonablypolymorphic.com/blog/higher-kinded-data/) to model our core `Term` data type. The goal was to be able to model validation, semi-complete data, and different models of required/optional in the same way. I think now that this approach was a red herring, since it was clear while modelling GraphQL that a simple `w` parameter in `Term` was not enough to model the required subtleties.

Right now, `w` is only set to `Identity` when the transport format is chosen to be Avro (since fields are required by default), and to `Maybe` when choosing Protocol Buffers (since all fields are optional). I think this is also a mistake, since:
1. It makes it very difficult to expose the same server between different transport mechanisms, which lead to mosntruosities like the `MaybeLike` type class in `mu-schema`.
2. The fact that fields are optional in Protocl Buffers should be reflected in the schema, and introduce was to fail in the conversion from Protocol Buffers to `Term` if a required element is not present.


serras:
@kutyel @diesalbla @pepegar @cb372 you might have an opinion here, so feel free to comment.  Overall, I think this makes the architecture much simpler and understandable to a newcomer.

cb372:
Nowadays I have a better understanding of the Protobuf spec and I can say it's actually a bit more complicated than "all fields are optional".

In Protobuf, every scalar type (string, bytes, int32, etc.) has a default value (`""`, `[]`, `0`, etc.). When writing, if a field is set to the default value for its type, the writer can (and should, to minimise payload size) skip that field and not write it to the wire. (The proto3 spec explains what to do for `oneof` fields, maps, repeated fields, etc.)

So when reading, if a field is missing, it means either:

1. The field is missing because the writer was using an old version of the protocol that doesn't have that field, or
2. The field is present but it has the default value for its type

It's impossible for the reader to distinguish between those cases.

In the Scala library we decided *not* to wrap scalar values in `Option` when generating code from `.proto`, because it seems a bit disingenuous in the face of this uncertainty. Instead we populate missing scalar fields with their default values when reading, and leave the user to decide what to do with them. See [these tests](https://github.com/47deg/pbdirect/blob/9202fea7b2f1d5ee8c3f3ff90454f53c343d8106/src/test/scala/pbdirect/PBFieldReaderSpec.scala#L158-L199) for examples. (It could be argued that wrapping all scalar fields in `Option` would make it more obvious that the field might be missing, and I've flip-flopped on this decision a number of times.)

For embedded message fields, there are 2 cases to consider:

1. the field is missing
2. the field is present but all of its fields are missing

This time, it *is* possible to distinguish between the cases because they are encoded differently on the wire. See [these tests](https://github.com/47deg/pbdirect/blob/9202fea7b2f1d5ee8c3f3ff90454f53c343d8106/src/test/scala/pbdirect/PBMessageWriterSpec.scala#L65-L82) for an example.

For this reason, we wrap embedded message fields in `Option` when we generate code from `.proto`.

However, our Protobuf serialization library can also handle embedded message fields *not* being optional - if a non-optional embedded message field is missing when deserializing, the library creates a value by populating all of its fields with their default values. See [this test](https://github.com/47deg/pbdirect/blob/9202fea7b2f1d5ee8c3f3ff90454f53c343d8106/src/test/scala/pbdirect/PBFieldReaderSpec.scala#L200-L205) for an example.

Hope this helps.

serras:
Thanks @cb372, this is incredibly useful! Indeed, it seems that I was too naïve on reading the spec as "all fields are optional". As part of this PR I'm going to implement the right behavior, since dropping `Maybe` everywhere forces me to choose one behavior, so it'd rather be the good one.

PR #131, opened 2020-03-09
Expose all GraphQL API as a WAI application

### LINKED ISSUE #121: GraphQL: expose as WAI Application
Connect the whole pipeline: get the query in a request, parse it using Hasura's parser, then turn that into a typed query, and finally run it against a resolver. We should try to expose it directly using WAI's API (so no Servant or Spock), to not have that many indirection layers.

*Sub-task: try to match both query and mutation when running the resolver.* We might need to introduce some annotations to mark which of the objects in the resolver should be thought of as the “query” one and which as the “mutation” one.

kutyel:
Fixes #121


graphql/src/Mu/GraphQL/Server.hs
```diff
@@ -0,0 +1,155 @@
...
+      toObjFld :: (T.Text, ValueConst) -> ObjectFieldG ValueConst
+      toObjFld (k, v) = ObjectFieldG (coerce k) v
+
+-- | Turn a Mu GraphQL 'Server' into a WAI 'Application'.
+--
+--   These 'Application's can be later combined using,
+--   for example, @wai-routes@, or you can add middleware
+--   from @wai-extra@, among others.
+graphQLApp ::
+    ( p ~ 'Package pname ss
```
serras: We need to create a type synonym for this monstruosity. Maybe?

```haskell
type GraphQLApp p pname ss hs chn qr qanns qmethods mut mannd mmethods = big thing
```

graphql/src/Mu/GraphQL/Server.hs
```diff
@@ -0,0 +1,155 @@
...
+          execQuery Nothing empty (decodeUtf8 $ toStrict body)
+        _                          -> toError "No `Content-Type` header found!"
+    _          -> toError "Unsupported method"
+  where
+    execQuery :: Maybe T.Text -> VariableMapC -> T.Text -> IO ResponseReceived
+    execQuery opn vals qry =
+      case parseExecutableDoc qry of
+        Left err  -> toError err
+        Right doc -> do
+          response <- runPipeline server q m opn vals doc
```
serras: `>>=`?

PR #134, opened 2020-03-12
Better parser errors for GraphQL documents

serras:


graphql/src/Mu/GraphQL/Server.hs
```diff
@@ -59,18 +57,6 @@ instance A.FromJSON ValueConst where
       toObjFld :: (T.Text, ValueConst) -> ObjectFieldG ValueConst
       toObjFld (k, v) = ObjectFieldG (coerce k) v
 
-type GraphQLApp p pname ss qmethods mmethods hs chn qr mut qanns manns =
```
kutyel: Since you moved this to another file, you can now safely remove `ConstraintKinds` on https://github.com/higherkindness/mu-haskell/pull/134/files#diff-c037c2c61f81962ca06b553051605319R1

graphql/src/Mu/GraphQL/Query/Parse.hs
```diff
@@ -184,192 +193,245 @@ instance ParseArgs p '[] where
...
 instance {-# OVERLAPPABLE #-} FindDefaultArgValue xs
          => FindDefaultArgValue (x ': xs) where
   findDefaultArgValue _ = findDefaultArgValue (Proxy @xs)
 instance {-# OVERLAPS #-} ReflectValueConst v
          => FindDefaultArgValue (DefaultValue v ': xs) where
-  findDefaultArgValue _ = pure $ reflectValueConst (Proxy @v)
+  findDefaultArgValue _ _ = pure $ reflectValueConst (Proxy @v)
 
-parseArg' :: (ParseArg p a, Alternative f)
+parseArg' :: (ParseArg p a, MonadError T.Text f)
```
kutyel: So basically when you want to fail with a message, you `throwError` and rely on `MonadError` instead of Alternative|Applicative? Seems useful! 🤔

serras: Yep, that's also why having `Alternative f` is useful, you just replace it by `MonadError e` and the compiler tells you where to change things.

One problem with this approach is that you can only throw *one* error for the entire parsing, whereas in some cases you may want to give as many errors as possible.

PR #139, opened 2020-03-12
Unified examples

### LINKED ISSUE #81: GraphQL: Convert gRPC <-> GraphQL
Figure out how we can directly expose a gRPC server as a GraphQL one with the minimal amount of boilerplate.

*Possible options:*
- **Preferred one:** do some type-level magic to turn the Schema into a bunch of Service’s, along with the corresponding “dumb handler” for each of them. Then turn the original Service into a new Service which refers to the former ones.
- **Another one**: read gRPC schemas following GraphQL semantics, and generate the “dumb handlers” using Template Haskell.

serras:
Fixes #81


serras:
We may need to rework this idea later, but right now "streams" in gRPC are mapped to "lists" in GraphQL.

There's another problem too: everything goes as query or mutation, the system does not really care. But if we have subscriptions, we might want to treat things returning "streams" differently from the rest.

PR #145, opened 2020-03-17
Add a CI job to test the Stack templates

cb372:
The first run takes about 15 minutes to download and compile the world, but subsequent runs will take about 2 minutes because `~/.stack` is cached.


cb372:
I can move it to that file, but are you sure you only want to find out about breakages after a change has been merged to master? Seems like a slightly unusual workflow to me.

kutyel:
Thanks Chris! We'll do a quick test of moving back to Stack for a couple commits, to so how it behaves (since haskell.nix has no support yet for the lts-15). Thanks again! 🙌

PR #146, opened 2020-03-18
Subscriptions

### LINKED ISSUE #82: GraphQL: Subscriptions
Add the ability to define subscription objects, and push changes via WebSockets.

*Notes about this task:*
- In the type-level we can have the subscription operations return a `Stream`. This makes the `ServerT` directly require a `Conduit` to push information to, which is great.
- The main issue here is that subscriptions are usually exposed via WebSockets. It seems that [`wai-websockets`](http://hackage.haskell.org/package/wai-websockets) provides a way to expose WebSockets as WAI Applications, but this requires some preliminary research.

serras:
Closes #82

### DIFF
--- a/graphql/exe/Main.hs
+++ b/graphql/exe/Main.hs
@@ -11,12 +11,16 @@
 
 module Main where
 
-import           Data.List              (find)
-import           Data.Maybe             (fromMaybe, listToMaybe)
+import           Data.List                         (find)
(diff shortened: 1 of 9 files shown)


serras:
Works with the [Python GraphQL client](https://pypi.org/project/py-graphql-client/) but not with GraphQL Playground 🤔

PR #151, opened 2020-03-20
Update to warp-grpc 0.4 and fix multi-servers

serras:
The initial goal was update to the new [`warp-grpc` 0.4](http://hackage.haskell.org/package/warp-grpc), which includes its own functionality to work with transformers.

While doing so, I discovered that the gRPC spec does not allow services to be served from routes, which means that the examples using `wai-route` didn't work *at all*. This PR also fixes this by running several applications concurrently on different ports.

### DIFF
--- a/examples/seed/src/Main.hs
+++ b/examples/seed/src/Main.hs
@@ -51,7 +51,7 @@ main = do
   putStrLn "running seed application"
   runConcurrently $ (\_ _ _ -> ())
     <$> Concurrently (runGRpcAppTrans msgProtoBuf 8080 runStderrLoggingT server)
-    <*> Concurrently (runGRpcAppTrans msgAvro     8080 runStderrLoggingT server)
+    <*> Concurrently (runGRpcAppTrans msgAvro     8081 runStderrLoggingT server)
     <*> Concurrently (runGraphQLAppTrans         50053 runStderrLoggingT server
                         (Proxy @('Just "PeopleService"))
                         (Proxy @'Nothing) (Proxy @'Nothing))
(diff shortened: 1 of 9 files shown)


examples/seed/src/Main.hs
```diff
@@ -50,14 +49,12 @@ newtype PeopleResponse = PeopleResponse
...
-    [ defRoute (str "proto" ./ end) $
-        \_ -> gRpcAppTrans msgProtoBuf runStderrLoggingT server
-    , defRoute (str "avro" ./ end) $
-        \_ -> gRpcAppTrans msgAvro runStderrLoggingT server
-    , defRoute (str "graphql" ./ end) $
-        \_ -> graphQLAppTransQuery runStderrLoggingT server (Proxy @"PeopleService")
-    ]
+  runConcurrently $ (\_ _ _ -> ())
+    <$> Concurrently (runGRpcAppTrans msgProtoBuf 8080 runStderrLoggingT server)
+    <*> Concurrently (runGRpcAppTrans msgAvro     8080 runStderrLoggingT server)
```
cb372: Shouldn't these port numbers be distinct?

cb372:
>gRPC spec does not allow services to be served from routes

It's fine to run multiple services on one server, even with different serialization formats per service, but yeah you can't stick `avro/` or `proto/` in front of the routes like you were doing in the examples.

cb372:
2 hour 41 minute build! 😎

serras:
Yes, every time we need to update the set of available packages the cache needs to be rebuilt 😞

PR #155, opened 2020-03-24
Experiment with Stack on CI 🧪 (lts-15)

kutyel:
Let's see how it goes with Stack 😄 
I also tried to cache the markdown files of the docs so maybe the deploy step should be faster too! 🚀

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,7 +12,7 @@ jobs:
     - uses: actions/cache@v1
       with:
         path: ~/.stack
-        key: dot-stack-${{ hashFiles('docs/**/*.md') }}
+        key: dot-stack-${{ hashFiles('stack.yaml') }}
         restore-keys: |
           dot-stack-
     - name: Get Standalone Haddock
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,6 +9,12 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - uses: mstksg/setup-stack@v1
+    - uses: actions/cache@v1
+      with:
+        path: ~/.stack
+        key: dot-stack-${{ hashFiles('docs/**/*.md') }}
+        restore-keys: |
+          dot-stack-
     - name: Get Standalone Haddock
       run: stack install standalone-haddock
     - name: Get rdkafka
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -16,26 +16,3 @@ jobs:
       run: sudo apt-get install librdkafka-dev
     - name: Compile with Stack
       run: stack build
-  test_stack_templates:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v1
-    - uses: mstksg/setup-stack@v1
-    - uses: actions/cache@v1
-      with:
-        path: ~/.stack
-        key: dot-stack-templates-${{ hashFiles('templates/*.hsfiles') }}
-        restore-keys: |
-          dot-stack-templates-
-    - name: Generate Avro RPC server from Stack template
-      run: stack new muavro $GITHUB_WORKSPACE/templates/grpc-server-avro.hsfiles -p "author-email:haskell.curry@47deg.com" -p "author-name:Haskell Curry"
-      working-directory: /tmp
-    - name: Build the generated Avro RPC server
-      run: stack build
-      working-directory: /tmp/muavro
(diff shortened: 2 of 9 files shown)


.github/workflows/haskell.yml
```diff
@@ -5,38 +5,25 @@ jobs:
...
-      - name: Build the generated Protobuf RPC server
-        run: stack build
-        working-directory: /tmp/muhaskellprotobuf
+        path: ~/.stack
+        key: dot-stack-${{ hashFiles('stack.yaml') }}
+        restore-keys: |
+          dot-stack-
+    - name: Compile with Stack
+      run: stack build
+    # Test Stack Templates
```
cb372: You don't want to run this as a separate job? I prefer to split the workflow into fine-grained jobs so that a) they run in parallel, and b) it's easy to see what failed without scrolling through 1000s of lines of logs. But it's up to you - it's your project :)

.github/workflows/deploy.yml
```diff
@@ -9,6 +9,12 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - uses: mstksg/setup-stack@v1
+    - uses: actions/cache@v1
+      with:
+        path: ~/.stack
+        key: dot-stack-${{ hashFiles('docs/**/*.md') }}
```
cb372: I'm a bit confused by this. Normally the cache key should be derived from something that will actually affect the contents of the cache. Do you expect a change in the markdown files to affect the contents of `~/.stack`?

.github/workflows/haskell.yml
kutyel: Makes sense, thanks! 😄

.github/workflows/deploy.yml
kutyel: You are right, but then the only sensible thing that can go there is `stack.yaml`! 😉

.github/workflows/haskell.yml
```diff
@@ -5,38 +5,37 @@ jobs:
...
-
-      - name: Build the generated Protobuf RPC server
-        run: stack build
-        working-directory: /tmp/muhaskellprotobuf
+    - uses: actions/checkout@v1
+    - uses: mstksg/setup-stack@v1
+    - uses: actions/cache@v1
+      with:
+        path: ~/.stack
+        key: dot-stack-${{ hashFiles('templates/*.hsfiles') }}
```
cb372: I think the caches for both jobs will be shared, because they are in the same workflow. So you might want to give them different names so they don't clash with each other?

If you give them a common prefix, e.g. `dot-stack-build-...` and `dot-stack-templates-...`, you can use that prefix as a fallback in `restore-keys`, so you should never end up with a completely cold cache:

```
restore-keys: |
  dot-stack-templates-
  dot-stack-
```

The action will try each prefix in turn until it finds a cache it can use. More details [here](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key).

kutyel:
We need a way to test the local templates on each commit, not with the URL to the repo, any ideas? 🤔

cb372:
https://docs.haskellstack.org/en/stable/GUIDE/#templates suggests it's possible to load a template from the filesystem. I think `stack new $GITHUB_WORKSPACE/mu-haskell/templates/grpc-server-avro.hsfiles` should work.

serras:
@cb372 after discussing it with @kutyel, we have decided to drop the template check from CI. The reason is that templates refer to released packages. However, in CI we would like to test the templates against the version in `master`, not the released ones.

Instead, in a PR yet to come, we are going to create a "release checklist", which includes checking not only the templates, but also running some other checks such the schema compatibility.

serras:
I ended up adding the `RELEASE.md` file here, it made no sense to merge this without that file

@rachelcarmena could you check that our release procedure looks right, especially in terms of using Git tags correctly? Thanks upfront :)

rachelcarmena:
> I ended up adding the `RELEASE.md` file here, it made no sense to merge this without that file
> 
> @rachelcarmena could you check that our release procedure looks right, especially in terms of using Git tags correctly? Thanks upfront :)

Thanks to you for your trust, Alejandro!! :raised_hands: It looks very good to me though it's easy to automate it (btw, `hub` works like a charm to create release notes from command line).

serras:
> Thanks to you for your trust, Alejandro!! 🙌 It looks very good to me though it's easy to automate it (btw, `hub` works like a charm to create release notes from command line).

We want to do that in the future, but right now the release procedure includes checking the output of some commands, something for which we haven't found (yet) a way to automate.

PR #156, opened 2020-03-24
Fix the documentation for the initGRpc function

cb372:


kutyel:
Our docs need a serious revamp after the latest changes to the API 😅

PR #160, opened 2020-03-25
Various fixes to GraphQL introspection

serras:


kutyel:
All the new builds are taking ~25m again 😢 I think we should go back to cachix! 🔥

serras:
What about not caching .stack-work? That used to work fine.

kutyel:
Looking at the Actions tab, none of the caches for stack seem to be working 🤷‍♂

serras:
Could you try pushing something that doesn't change stack.yaml in this
branch to see whether it gets the right one?

kutyel:
Now it worked, could it be that the cache is not persisted between PRs? That's a really strange behaviour from the Github Actions Cache plugin! 🤔

serras:
My guess is that we were making a lot of changes in the stack.yaml files
lately, and that from now on it will stabilize.

PR #166, opened 2020-03-30
Read GraphQL Schema Definitions

### LINKED ISSUE #83: GraphQL: Read GraphQL schema definitions
Turn them into the corresponding Schema and set of Service’s. The implementation should be quite similar to what we already do in Avro and Protocol Buffers quasi-quoters.

kutyel:
Fixes #83

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -20,7 +20,7 @@ jobs:
         key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ '**/*.hs' }}
         restore-keys: |
           ${{ runner.os }}-stack-work-
-    - name: Get rdkafka
-      run: sudo apt-get install librdkafka-dev
+    # - name: Get rdkafka
+    #   run: sudo apt-get install librdkafka-dev
     - name: Compile with Stack
       run: stack build
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -17,7 +17,7 @@ jobs:
       name: Cache .stack-work
       with:
         path: .stack-work
-        key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ hashFiles('**/*.hs') }}
+        key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ '**/*.hs' }}
         restore-keys: |
           ${{ runner.os }}-stack-work-
     - name: Get rdkafka
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -17,10 +17,10 @@ jobs:
       name: Cache .stack-work
       with:
         path: .stack-work
-        key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ '**/*.hs' }}
+        key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ hashFiles('**/*.hs') }}
         restore-keys: |
           ${{ runner.os }}-stack-work-
-    # - name: Get rdkafka
-    #   run: sudo apt-get install librdkafka-dev
+    - name: Get rdkafka
+      run: sudo apt-get install librdkafka-dev
     - name: Compile with Stack
       run: stack build
--- a/graphql/mu-graphql.cabal
+++ b/graphql/mu-graphql.cabal
@@ -48,6 +48,7 @@ library
     , template-haskell
     , text
     , unordered-containers
+    , uuid
     , wai
     , wai-websockets
     , warp
--- a/graphql/mu-graphql.cabal
+++ b/graphql/mu-graphql.cabal
@@ -21,6 +21,7 @@ library
     Mu.GraphQL.Server
 
   other-modules:
+    Mu.GraphQl.Quasi.LostParser
     Mu.GraphQL.Query.Definition
     Mu.GraphQL.Query.Introspection
     Mu.GraphQL.Query.Parse
@@ -31,6 +32,7 @@ library
   build-depends:
       aeson
     , async
+    , attoparsec
     , base                  >=4.12 && <5
     , bytestring
     , conduit
@@ -40,6 +42,7 @@ library
     , mtl
     , mu-rpc
     , mu-schema
+    , parsers
     , scientific
     , sop-core
     , stm
--- a/graphql/mu-graphql.cabal
+++ b/graphql/mu-graphql.cabal
@@ -21,7 +21,7 @@ library
     Mu.GraphQL.Server
 
   other-modules:
-    Mu.GraphQl.Quasi.LostParser
+    Mu.GraphQL.Quasi.LostParser
     Mu.GraphQL.Query.Definition
     Mu.GraphQL.Query.Introspection
     Mu.GraphQL.Query.Parse
--- a/graphql/mu-graphql.cabal
+++ b/graphql/mu-graphql.cabal
@@ -12,6 +12,7 @@ category:      Network
 build-type:    Simple
 homepage:      https://higherkindness.io/mu-haskell/
 bug-reports:   https://github.com/higherkindness/mu-haskell/issues
+data-files:    exe/*.graphql
 
 library
   exposed-modules:
--- a/graphql/mu-graphql.cabal
+++ b/graphql/mu-graphql.cabal
@@ -16,6 +16,7 @@ bug-reports:   https://github.com/higherkindness/mu-haskell/issues
 library
   exposed-modules:
     Mu.GraphQL.Annotations
+    Mu.GraphQL.Quasi
     Mu.GraphQL.Server
 
   other-modules:
@@ -44,6 +45,7 @@ library
     , stm-chans
     , stm-conduit
     , stm-containers
+    , template-haskell
     , text
     , unordered-containers
     , wai
@@ -62,12 +64,12 @@ executable library-graphql
   default-language: Haskell2010
   ghc-options:      -Wall
   build-depends:
-      base             >=4.12   && <5
+      base        >=4.12    && <5
     , conduit
     , mu-graphql
     , mu-rpc
     , mu-schema
-    , regex-tdfa       >=1.3.1.0
+    , regex-tdfa  >=1.3.1.0
     , text
-    , warp
     , wai-extra
+    , warp
--- a/graphql/src/Mu/GraphQL/Quasi.hs
+++ b/graphql/src/Mu/GraphQL/Quasi.hs
@@ -4,6 +4,7 @@
 {-# language ViewPatterns      #-}
 
 module Mu.GraphQL.Quasi (
+  -- * 'Schema' generation from @.graphql@ files
   graphql
 ) where
 
--- a/graphql/src/Mu/GraphQL/Quasi.hs
+++ b/graphql/src/Mu/GraphQL/Quasi.hs
@@ -5,7 +5,9 @@
 
 module Mu.GraphQL.Quasi where
 
+import           Data.Int                      (Int32)
 import qualified Data.Text                     as T
+import           Data.UUID                     (UUID)
 import           Language.GraphQL.Draft.Parser (parseSchemaDoc)
 import qualified Language.GraphQL.Draft.Syntax as GQL
 import           Language.Haskell.TH
@@ -38,7 +40,15 @@ graphqlToDecls schemaName serviceName (GQL.SchemaDocument types) = do
   pure [schemaDec, serviceDec]
 
 typeToDec :: Name -> GQL.TypeDefinition -> Q Result
-typeToDec _ (GQL.TypeDefinitionScalar _)             = error "not implemented" -- TODO: handle "well-known scalars"
+typeToDec _ (GQL.TypeDefinitionScalar (GQL.ScalarTypeDefinition _ s _)) = GQLSchema <$> scalarToType s
+  where
+    scalarToType :: GQL.Name -> Q Type
+    scalarToType (GQL.unName -> "Int")     = [t|'TPrimitive Int32|]
+    scalarToType (GQL.unName -> "Float")   = [t|'TPrimitive Double|]
+    scalarToType (GQL.unName -> "String")  = [t|'TPrimitive T.Text|]
+    scalarToType (GQL.unName -> "Boolean") = [t|'TPrimitive Bool|]
+    scalarToType (GQL.unName -> "ID")      = [t|'TPrimitive UUID|]
+    scalarToType _                         = fail "not well-known scalar types are not supported"
 typeToDec _ (GQL.TypeDefinitionObject objs) = objToDec objs
   where
     objToDec :: GQL.ObjectTypeDefinition -> Q Result
--- a/graphql/src/Mu/GraphQL/Quasi.hs
+++ b/graphql/src/Mu/GraphQL/Quasi.hs
@@ -5,8 +5,10 @@
 
 module Mu.GraphQL.Quasi where
 
+import           Control.Monad.IO.Class        (liftIO)
 import           Data.Int                      (Int32)
 import qualified Data.Text                     as T
+import qualified Data.Text.IO                  as TIO
 import           Data.UUID                     (UUID)
 import           Language.GraphQL.Draft.Parser (parseSchemaDoc)
 import qualified Language.GraphQL.Draft.Syntax as GQL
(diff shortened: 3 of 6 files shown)


graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,154 @@
...
+      serviceName' = mkName serviceName
+  rs <- traverse (typeToDec schemaName') types
+  let schemaTypes  = [x | GQLSchema  x <- rs]
+      serviceTypes = [x | GQLService x <- rs]
+  schemaDec <- tySynD schemaName' [] (pure $ typesToList schemaTypes)
+  serviceDec <- tySynD serviceName' [] (pure $ typesToList serviceTypes)
+  pure [schemaDec, serviceDec]
+
+typeToDec :: Name -> GQL.TypeDefinition -> Q Result
+typeToDec _ (GQL.TypeDefinitionScalar _)          = fail "scalar types are not supported"
```
serras, line 40: Remember that here you need to include special cases for the "well-known" scalar types, not fail right away.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,154 @@
...
+typeToDec schemaName (GQL.TypeDefinitionObject objs)       = objToDec objs
+  where
+    objToDec :: GQL.ObjectTypeDefinition -> Q Result
+    objToDec (GQL.ObjectTypeDefinition _ nm _ _ flds) =
+      GQLService <$> [t| 'Service $(textToStrLit $ GQL.unName nm) '[]
+          $(typesToList <$> traverse gqlFieldToType flds) |]
+    gqlFieldToType :: GQL.FieldDefinition -> Q Type
+    gqlFieldToType (GQL.FieldDefinition _ fnm args ftyp _) =
+      [t| 'Method $(textToStrLit $ GQL.unName fnm) '[]
```
serras, line 49: Indentation is slightly odd in your splices.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,154 @@
...
+typeToDec :: Name -> GQL.TypeDefinition -> Q Result
+typeToDec _ (GQL.TypeDefinitionScalar _)          = fail "scalar types are not supported"
+typeToDec schemaName (GQL.TypeDefinitionObject objs)       = objToDec objs
+  where
+    objToDec :: GQL.ObjectTypeDefinition -> Q Result
+    objToDec (GQL.ObjectTypeDefinition _ nm _ _ flds) =
+      GQLService <$> [t| 'Service $(textToStrLit $ GQL.unName nm) '[]
+          $(typesToList <$> traverse gqlFieldToType flds) |]
+    gqlFieldToType :: GQL.FieldDefinition -> Q Type
+    gqlFieldToType (GQL.FieldDefinition _ fnm args ftyp _) =
```
serras, line 48: May I suggest using a view pattern on `fnm` to make the code below slightly less verbose?

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,154 @@
...
+      [t| 'Method $(textToStrLit $ GQL.unName fnm) '[]
+        $(typesToList <$> traverse argToType args)
+        $(retToType ftyp) |]
+    argToType :: GQL.InputValueDefinition -> Q Type
+    argToType (GQL.InputValueDefinition _ aname _ _) =
+      [t| 'ArgSingle 'Nothing '[] ('SchemaRef $(conT schemaName) $(textToStrLit (GQL.unName aname))) |]
+    retToType :: GQL.GType -> Q Type
+    retToType (GQL.TypeNamed _ a) =
+      [t| 'RetSingle ('SchemaRef $(conT schemaName) $(textToStrLit . GQL.unName . GQL.unNamedType $ a)) |]
+    retToType _ = fail "only named types may be used as results"
```
serras, line 58: That's not true! You can also return lists of stuff. Furthermore, you need to check the nullability field to see whether you need to wrap everything on a `OptionalRef` or not.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,154 @@
...
+          $(typesToList <$> traverse gqlFieldToType flds) |]
+    gqlFieldToType :: GQL.FieldDefinition -> Q Type
+    gqlFieldToType (GQL.FieldDefinition _ fnm args ftyp _) =
+      [t| 'Method $(textToStrLit $ GQL.unName fnm) '[]
+        $(typesToList <$> traverse argToType args)
+        $(retToType ftyp) |]
+    argToType :: GQL.InputValueDefinition -> Q Type
+    argToType (GQL.InputValueDefinition _ aname _ _) =
+      [t| 'ArgSingle 'Nothing '[] ('SchemaRef $(conT schemaName) $(textToStrLit (GQL.unName aname))) |]
+    retToType :: GQL.GType -> Q Type
```
serras, line 55: For later it's important to know that when you are dealing with the subscription type (which you have to detect from the schema), you should not use `RetSingle` but `RetStream`.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,154 @@
...
+--   { _iotdDescription      :: !(Maybe Description)
+--   , _iotdName             :: !Name
+--   , _iotdDirectives       :: ![Directive]
+--   , _iotdValueDefinitions :: ![InputValueDefinition]
+--   }
+
+schemaFromGQL :: [GQL.ValueConst] -> Q Type
+schemaFromGQL = (typesToList <$>) . traverse schemaFromGQLValue
+
+schemaFromGQLValue :: GQL.ValueConst -> Q Type
```
serras, line 130: This is not useful at all. You will need to eventually write a function with that signature, but not to return the type, but to handle default values. For now on, you can just remove it.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,148 @@
...
+      GQLService <$> [t| 'Service $(textToStrLit $ GQL.unName nm) '[]
+          $(typesToList <$> traverse gqlFieldToType flds) |]
+    gqlFieldToType :: GQL.FieldDefinition -> Q Type
+    gqlFieldToType (GQL.FieldDefinition _ (GQL.unName -> fnm) args ftyp _) =
+      [t| 'Method $(textToStrLit fnm) '[]
+            $(typesToList <$> traverse argToType args)
+            'RetSingle $(retToType ftyp) |]
+    argToType :: GQL.InputValueDefinition -> Q Type
+    argToType (GQL.InputValueDefinition _ aname _ _) =
+      [t| 'ArgSingle 'Nothing '[] ('SchemaRef $(conT schemaName) $(textToStrLit (GQL.unName aname))) |]
```
serras, line 54: This is not correct, in several respects:

- if `aname` is the name of the argument, it should go as `Just $(textToStrLit (GQL.unName aname))` instad of `Nothing`,
- the last part should be the **type** of the argument. But it cannot be always `SchemaRef`, you have to take into account that it could be a list (to wrap in a `ListRef`), whether it's optional (to use `OptionalRef`), or whether it's a scalar (to use `TPrimitive` or give an error).

kutyel: I think I've fixed that, but now I don't have any reference to `'SchemaRef schemaName` 🤔 where should that happen?

.github/workflows/haskell.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       name: Cache .stack-work
       with:
         path: .stack-work
-        key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ '**/*.hs' }}
+        key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ hashFiles('**/*.hs') }}
```
serras, line 20: I am not super happy about slipping changes about other stuff in the same PR.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,106 @@
...
+    inputObjToDec :: GQL.InputObjectTypeDefinition -> Q Result
+    inputObjToDec (GQL.InputObjectTypeDefinition _ (GQL.unName -> name) _ fields) =
+        GQLSchema <$> [t|'DRecord $(textToStrLit name)
+                                  $(typesToList <$> traverse gqlFieldToType fields)|]
+    gqlFieldToType :: GQL.InputValueDefinition -> Q Type
+    gqlFieldToType (GQL.InputValueDefinition _ (GQL.unName -> fname) ftype _) =
+      [t|'FieldDef $(textToStrLit fname) $(gtypeToType ftype)|]
+
+gtypeToType :: GQL.GType -> Q Type
+gtypeToType (GQL.TypeNamed (GQL.unNullability -> False) (GQL.unName . GQL.unNamedType -> a)) =
```
serras, line 92: This is not right, since the type could also be a scalar. In that case you need to use `PrimitiveRef`.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,106 @@
...
+    gqlFieldToType :: GQL.FieldDefinition -> Q Type
+    gqlFieldToType (GQL.FieldDefinition _ (GQL.unName -> fnm) args ftyp _) =
+      [t| 'Method $(textToStrLit fnm) '[]
+            $(typesToList <$> traverse argToType args)
+            'RetSingle $(gtypeToType ftyp) |]
+    argToType :: GQL.InputValueDefinition -> Q Type
+    argToType (GQL.InputValueDefinition _ (GQL.unName -> aname) atype Nothing) =
+      [t| 'ArgSingle ('Just $(textToStrLit aname)) '[] $(gtypeToType atype) |]
+    argToType (GQL.InputValueDefinition _ (GQL.unName -> aname) atype (Just defs)) =
+      [t| 'ArgSingle ('Just $(textToStrLit aname)) '[] {-$(fromGQLValueConst defs)-} $(gtypeToType atype) |]
```
serras, line 69: The default value should go as an annotation `DefaultValue`, in the list which is the second argument. [This line](https://github.com/higherkindness/mu-haskell/blob/master/graphql/exe/Main.hs#L59) has an example of how to use it.

graphql/src/Mu/GraphQL/Quasi.hs
serras: [This](https://github.com/higherkindness/mu-haskell/blob/master/graphql/exe/Main.hs#L59) is an example of an object returning a primitive instead of an object reference.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,106 @@
...
+  rs <- traverse (typeToDec schemaName') types
+  let schemaTypes  = [x | GQLSchema  x <- rs]
+      serviceTypes = [x | GQLService x <- rs]
+  schemaDec <- tySynD schemaName' [] (pure $ typesToList schemaTypes)
+  serviceDec <- tySynD serviceName' [] (pure $ typesToList serviceTypes)
+  pure [schemaDec, serviceDec]
+
+-- | Reads a GraphQL 'TypeDefinition' and returns a 'Result'.
+typeToDec :: Name -> GQL.TypeDefinition -> Q Result
+typeToDec _ (GQL.TypeDefinitionScalar (GQL.ScalarTypeDefinition _ s _)) = GQLSchema <$> scalarToType s
```
serras, line 45: I don't understand what this is doing. Scalars do not go into either the schema or the service definition (you can return `Maybe Result` here and then use `catMaybes` or `mapMaybe` in `graphqlToDecs`).

What you generate here is more like the thing that would go into an argument or field definition.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,106 @@
...
+      [t|'ChoiceDef $(textToStrLit c)|]
+typeToDec _ (GQL.TypeDefinitionInputObject inpts) = inputObjToDec inpts
+  where
+    inputObjToDec :: GQL.InputObjectTypeDefinition -> Q Result
+    inputObjToDec (GQL.InputObjectTypeDefinition _ (GQL.unName -> name) _ fields) =
+        GQLSchema <$> [t|'DRecord $(textToStrLit name)
+                                  $(typesToList <$> traverse gqlFieldToType fields)|]
+    gqlFieldToType :: GQL.InputValueDefinition -> Q Type
+    gqlFieldToType (GQL.InputValueDefinition _ (GQL.unName -> fname) ftype _) =
+      [t|'FieldDef $(textToStrLit fname) $(gtypeToType ftype)|]
```
serras, line 89: You cannot use the same `gtypeToType` for both input objects and normal objects. In the latter case you do it correctly, but in the former you need to generate `TPrimitive`, `TSchematic`, and so on.

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,166 @@
...
+      serviceTypes = [x | GQLService x <- rs]
+  schemaDec <- tySynD schemaName' [] (pure $ typesToList schemaTypes)
+  serviceDec <- tySynD serviceName' []
+    [t| 'Package ('Just $(textToStrLit $ T.pack serviceName))
+                  $(pure $ typesToList serviceTypes) |]
+  pure [schemaDec, serviceDec]
+
+-- | Reads a GraphQL 'TypeDefinition' and returns a 'Result'.
+typeToDec :: Name -> TypeMap -> GQL.TypeDefinition -> Q Result
+typeToDec schemaName tm (GQL.TypeDefinitionScalar (GQL.ScalarTypeDefinition _ s _)) = GQLScalar <$ scalarToType s tm schemaName
```
serras, line 72: Line break?

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,166 @@
...
+    typeToPrimType (GQL.unName -> "ID")      = [t|'TPrimitive UUID|]
+    typeToPrimType (coerce -> name)          = [t|'TSchematic $(textToStrLit name)|]
+
+scalarToType :: GQL.Name -> TypeMap -> Name -> Q Type
+scalarToType (GQL.unName -> "Int") _  _     = [t|'PrimitiveRef Integer|]
+scalarToType (GQL.unName -> "Float") _ _    = [t|'PrimitiveRef Double|]
+scalarToType (GQL.unName -> "String") _ _   = [t|'PrimitiveRef T.Text|]
+scalarToType (GQL.unName -> "Boolean") _ _  = [t|'PrimitiveRef Bool|]
+scalarToType (GQL.unName -> "ID") _ _       = [t|'PrimitiveRef UUID|]
+scalarToType (coerce -> name) tm schemaName =
```
serras, line 155: The name `scalarToType` suggests that if the type is not a scalar you should fail instead. Maybe rename the function?

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,166 @@
...
+  let schemaRef = [t|'SchemaRef $(conT schemaName) $(textToStrLit name)|]
+   in case HM.lookup name tm of
+        Just Enum        -> schemaRef
+        Just InputObject -> schemaRef
+        _                -> [t|'ObjectRef $(textToStrLit name)|]
+
+typesToList :: [Type] -> Type
```
serras, line 162: We are going to end up writing a small library with all of these utilities...

graphql/mu-graphql.cabal
```diff
@@ -12,13 +12,16 @@ category:      Network
 build-type:    Simple
 homepage:      https://higherkindness.io/mu-haskell/
 bug-reports:   https://github.com/higherkindness/mu-haskell/issues
+data-files:    exe/*.graphql
```
serras, line 15: TIL! I didn't know you could use wildcards there!

graphql/src/Mu/GraphQL/Quasi.hs
```diff
@@ -0,0 +1,187 @@
...
+
+data GQLType =
+    Enum
+  | Object
+  | Scalar
+  | InputObject
+  | Other
+
+classifySchema :: [GQL.SchemaDefinition] -> SchemaMap
+classifySchema = HM.fromList . concatMap schemaToMap
```
serras, line 49: The code is OK, but we aware that unless GHC does extreme optimization, you are here generating an intermediate list which is them discarded. I would suggest to use `foldl'`, using the `HashMap`'s `insert` at each step. On the other hand, now you can use `concatMap`, which is quite useful...

graphql/src/Mu/GraphQL/Quasi/LostParser.hs
```diff
@@ -0,0 +1,38 @@
...
+
+rootOperationParser :: Parser GQL.RootOperationTypeDefinition
+rootOperationParser =
+  GQL.RootOperationTypeDefinition
+    <$> (operationTypeParser <* symbol ":")
+    <*> (GQL.NamedType <$> nameParser)
+
+operationTypeParser :: Parser GQL.OperationType
+operationTypeParser =
+  GQL.OperationTypeQuery <$ symbol "query"
```
serras, line 28: Wouldn't it be nice to have everything aligned? 😝

kutyel:
I also proposed the `LostParser`™️ to Hasura here: https://github.com/hasura/graphql-parser-hs/pull/24 in case they want to add it and we can remove it! 😉

PR #170, opened 2020-03-31
Packed fields in Protocol Buffers

### LINKED ISSUE #163: Protobuf: decoding a packed repeated enum field
Something I picked up in the mu-scala integration tests I'm writing.

```proto
syntax = "proto3";

package integrationtest;

message GetForecastRequest {
  string city = 1;
  reserved 2 to 4;
  int32 days_required = 5;
}

message GetForecastResponse {
  enum Weather {
    SUNNY = 0;
    CLOUDY = 1;
    RAINY = 2;
  }

  string last_updated = 1;
  repeated Weather daily_forecasts = 2;
}

service WeatherService {
  rpc getForecast(GetForecastRequest) returns (GetForecastResponse);
}
```

The mu-scala server encodes the `daily_forecasts` repeated enum field using [packed encoding](https://developers.google.com/protocol-buffers/docs/encoding#packed). It appears the mu-haskell client cannot decode it. The field gets set to `Nothing`.

If I tweak things on the Scala side to stop using packed encoding, the mu-haskell client can decode the field just fine.

I've [confirmed](https://github.com/47degrees/pbdirect/pull/66) that the official protoc CLI tool also encodes repeated enum values using packed encoding. So it seems like mu-scala is doing the right thing and mu-haskell should be able to decode it.

I guess the fix for this would be in [protoc-wire](https://github.com/awakesecurity/proto3-wire/blob/master/src/Proto3/Wire/Decode.hs)?

serras:
Fixes #163 @cb372 

The patch works by adding a new `supportsPacking` flag to types. If this is true, they should provide a way to parse a list of those things at once. The parser of `TList a` is aware of that, and tries to call it if possible.

### DIFF
--- a/adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
+++ b/adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
@@ -175,6 +175,10 @@ class ProtoBridgeTerm (sch :: Schema tn fn) (t :: TypeDef tn fn) where
 class ProtoBridgeEmbedTerm (sch :: Schema tn fn) (t :: TypeDef tn fn) where
   termToEmbedProto :: FieldNumber -> Term sch t -> PBEnc.MessageBuilder
   embedProtoToOneFieldValue :: PBDec.Parser PBDec.RawPrimitive (Term sch t)
+  -- support for packed encodings
+  -- https://developers.google.com/protocol-buffers/docs/encoding#packed
+  supportsPackingTerm :: Proxy (Term sch t) -> Bool
+  embedProtoToPackedFieldValue :: PBDec.Parser PBDec.RawPrimitive [Term sch t]
 
 class ProtoBridgeField (sch :: Schema tn fn) (ty :: tn) (f :: FieldDef tn fn) where
   fieldToProto :: Field sch f -> PBEnc.MessageBuilder
@@ -221,6 +225,8 @@ instance ProtoBridgeTerm sch ('DRecord name args)
          => ProtoBridgeEmbedTerm sch ('DRecord name args) where
   termToEmbedProto fid v = PBEnc.embedded fid (termToProto v)
   embedProtoToOneFieldValue = PBDec.embedded' (protoToTerm @_ @_ @sch @('DRecord name args))
+  supportsPackingTerm _ = False
+  embedProtoToPackedFieldValue = error "this is a bug, since we declare we do not support packed encoding"
 
 -- ENUMERATIONS
 -- ------------
@@ -233,8 +239,10 @@ instance TypeError ('Text "protobuf requires wrapping enums in a message")
 instance ProtoBridgeEnum sch name choices
          => ProtoBridgeEmbedTerm sch ('DEnum name choices) where
   termToEmbedProto fid (TEnum v) = enumToProto @_ @_ @sch @name fid v
-  embedProtoToOneFieldValue = do n <- PBDec.int32
-                                 TEnum <$> protoToEnum @_ @_ @sch @name n
+  embedProtoToOneFieldValue = PBDec.int32 >>= fmap TEnum . protoToEnum @_ @_ @sch @name
+  supportsPackingTerm _ = True
+  embedProtoToPackedFieldValue =
+    PBDec.packedVarints >>= traverse (fmap TEnum . protoToEnum @_ @_ @sch @name)
 
 class ProtoBridgeEnum (sch :: Schema tn fn) (ty :: tn) (choices :: [ChoiceDef fn]) where
   enumToProto :: FieldNumber -> NS Proxy choices -> PBEnc.MessageBuilder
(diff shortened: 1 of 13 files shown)


adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
```diff
@@ -287,8 +291,13 @@ instance {-# OVERLAPS #-}
          => ProtoBridgeField sch ty ('FieldDef name ('TList t)) where
   fieldToProto (Field (FList xs))  = foldMap (oneFieldValueToProto fieldId) xs
     where fieldId = fromInteger $ natVal (Proxy @(FindProtoBufId sch ty name))
-  protoToField = Field . FList <$> (PBDec.repeated protoToOneFieldValue `at` fieldId <|> pure [])
+  protoToField = Field . FList <$> go
     where fieldId = fromInteger $ natVal (Proxy @(FindProtoBufId sch ty name))
+          base = PBDec.repeated protoToOneFieldValue `at` fieldId <|> pure []
+          go | supportsPacking (Proxy @(FieldValue sch t))
+             = PBDec.one protoToPackedFieldValue [] `at` fieldId <|> base
```
cb372, line 298: So this will attempt to decode it as a packed field first, and fallback to decoding as unpacked if that fails?

I think that's fine, but if you have access to the protobuf wire type, you could also switch based on that. If it's `length_delimited`, it's a packed field. That's what we do in the Scala implementation: https://github.com/47degrees/pbdirect/blob/185c2f5cf76ecd81c6ce5269fb6e6ab49cc657de/src/main/scala/pbdirect/PBFieldReader.scala#L25-L26

adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
```diff
@@ -287,8 +291,13 @@ instance {-# OVERLAPS #-}
          => ProtoBridgeField sch ty ('FieldDef name ('TList t)) where
   fieldToProto (Field (FList xs))  = foldMap (oneFieldValueToProto fieldId) xs
```
cb372, line 292: You could use the same `supportsPacking` flag to write using packed encoding as well. But you have to make sure not to use packed encoding if the field has been annotated with `[packed=false]` in the `.proto` file. I'm not sure if you have that information available at this point.

serras: In theory we could add a new kind of annotation about whether they should be packed or not when serializing to Protocol Buffers.

adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
serras: Looking at the code in [`runGetPacked`](http://hackage.haskell.org/package/proto3-wire-1.1.0/docs/src/Proto3.Wire.Decode.html#runGetPacked), that's sort of what's happening. The `runGetPacked` function checks whether we are in the "length delimited" case, and if not it fails. This failure is then taken over by the `(<|>)`, which tries the other option.

adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
serras: Done! The last commit adds support for packed fields, and reads that information from the `.proto` file. If not found, it defaults to packed fields if the type supports that functionality.

adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
```diff
@@ -61,7 +61,8 @@ instance ProtoEnum Bool
 -- | Annotations for Protocol Buffers fields.
 data ProtoBufAnnotation
   = -- | Numeric field identifier for normal fields
-    ProtoBufId Nat
+    --   and whether it should be packed (only used for lists of number-like values)
+    ProtoBufId Nat Bool
     -- | List of identifiers for fields which contain a union
   | ProtoBufOneOfIds [Nat]
```
cb372, line 67: Should this be changed to `ProtoBufOneOfIds [ProtobufId]`?

e.g. in

```proto
oneof string_or_int {
  string b = 2;
  int32 c   = 3;
}
```

one choice can be packed, the other cannot.

serras: Oh, I sort of assumed that if you have a list of `oneof`, then that cannot be packed. So we can have a list of `string_or_int` and have it packed? But how do you know then which element of the union you chose in each case?

cb372: >A oneof cannot be repeated.

Huh! Didn't know that. So you can ignore my comment.

https://developers.google.com/protocol-buffers/docs/proto3#oneof-features

serras: One more argument for my "Protocol Buffers has the most subtleties ever" rant, to appear shortly xD

PR #173, opened 2020-04-07
Better API to declare servers

### LINKED ISSUE #136: Better notation for building servers
Right now, building a server requires matching the exact order in which methods and services are defined in the correspoding type. When this type comes from importing a `.proto`, `.avdl`, or in the future GraphQL schema, keeping this order is quite difficult. Furthermore, @kutyel has noticed that the error message you get is quite un-informative.

So we need to find a better notation. Right now we are thinking of a `blaze-builder`-like notation using `do`, like:

```haskell
service @"Book" $ do
  method @"id" ...
  method @"author" ...
```

@arianvp you might be interested in this

serras:
Fixes #136 

This defines a new API to build servers which does not rely on the order in which methods are declared in the type-level schema. This is mostly problematic because we expect people to *import* the schema, and the parser may not guarantee ordering (what happens if you have an import, for example?). 

```diff
quickstartServer
-  -- = Server (sayHello :<|>: sayHi :<|>: sayManyHellos :<|>: H0)
+  = singleService ( method @"SayHello" sayHello
+                  , method @"SayManyHellos" sayManyHellos
+                  , method @"SayHi" sayHi )
```

### DIFF
--- a/core/rpc/src/Mu/Rpc/Examples.hs
+++ b/core/rpc/src/Mu/Rpc/Examples.hs
@@ -9,6 +9,7 @@
 {-# language PartialTypeSignatures #-}
 {-# language PolyKinds             #-}
 {-# language ScopedTypeVariables   #-}
+{-# language TypeApplications      #-}
 {-# language TypeFamilies          #-}
 {-# language TypeOperators         #-}
 {-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
@@ -71,7 +72,11 @@ newtype HiRequest = HiRequest { number :: Int }
 quickstartServer :: forall m. (MonadServer m)
                  => ServerT '[] QuickStartService m _
 quickstartServer
-  = Server (sayHello :<|>: sayHi :<|>: sayManyHellos :<|>: H0)
+  -- = Server (sayHello :<|>: sayHi :<|>: sayManyHellos :<|>: H0)
+  = singleService $ method @"SayHello" sayHello
+                :|: method @"SayManyHellos" sayManyHellos
+                :|: method @"SayHi" sayHi
+                :|: N0
   where
     sayHello :: HelloRequest -> m HelloResponse
     sayHello (HelloRequest nm)
--- a/core/rpc/src/Mu/Rpc/Examples.hs
+++ b/core/rpc/src/Mu/Rpc/Examples.hs
@@ -91,20 +91,7 @@ quickstartServer
     sayManyHellos source sink
       = runConduit $ source .| C.mapM sayHello .| sink
 
-{-
-From https://www.apollographql.com/docs/apollo-server/schema/schema/
-
-type Book {
-  title: String
-  author: Author
-}
-
-type Author {
-  name: String
-  books: [Book]
-}
--}
-
+-- From https://www.apollographql.com/docs/apollo-server/schema/schema/
 type ApolloService
   = 'Package ('Just "apollo")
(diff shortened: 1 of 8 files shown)


serras:
Pinging @arianvp

core/rpc/src/Mu/Rpc/Examples.hs
```diff
@@ -120,9 +111,11 @@ type ApolloBookAuthor = '[
...
 apolloServer :: forall m. (MonadServer m) => ServerT ApolloBookAuthor ApolloService m _
 apolloServer
-  = Services $ (pure . fst :<||>: pure . snd :<||>: H0)
-               :<&>: (authorName :<||>: authorBooks :<||>: H0)
-               :<&>: S0
+  = resolver
+      ( object @"Author" ( field @"name"   authorName
+                         , field @"books"  authorBooks )
+      , object @"Book"   ( field @"author" (pure . snd)
+                         , field @"title"  (pure . fst) ) )
```
kutyel, line 118: Much better! 😍

core/rpc/src/Mu/Server.hs
```diff
@@ -207,3 +212,114 @@ instance (MonadError ServerError m, ToRef chn ref v, handler ~ m v)
...
+  toNamedList (n1, n2, n3, n4, n5, n6) = n1 :|: n2 :|: n3 :|: n4 :|: n5 :|: n6 :|: N0
+instance ToNamedList (Named n1 h1, Named n2 h2, Named n3 h3, Named n4 h4, Named n5 h5, Named n6 h6, Named n7 h7)
+                     '[ '(n1, h1), '(n2, h2), '(n3, h3), '(n4, h4), '(n5, h5), '(n6, h6), '(n7, h7) ] where
+  toNamedList (n1, n2, n3, n4, n5, n6, n7) = n1 :|: n2 :|: n3 :|: n4 :|: n5 :|: n6 :|: n7 :|: N0
+instance ToNamedList (Named n1 h1, Named n2 h2, Named n3 h3, Named n4 h4, Named n5 h5, Named n6 h6, Named n7 h7, Named n8 h8)
+                     '[ '(n1, h1), '(n2, h2), '(n3, h3), '(n4, h4), '(n5, h5), '(n6, h6), '(n7, h7), '(n8, h8) ] where
+  toNamedList (n1, n2, n3, n4, n5, n6, n7, n8) = n1 :|: n2 :|: n3 :|: n4 :|: n5 :|: n6 :|: n7 :|: n8 :|: N0
+instance ToNamedList (Named n1 h1, Named n2 h2, Named n3 h3, Named n4 h4, Named n5 h5, Named n6 h6, Named n7 h7, Named n8 h8, Named n9 h9)
+                     '[ '(n1, h1), '(n2, h2), '(n3, h3), '(n4, h4), '(n5, h5), '(n6, h6), '(n7, h7), '(n8, h8), '(n9, h9) ] where
+  toNamedList (n1, n2, n3, n4, n5, n6, n7, n8, n9) = n1 :|: n2 :|: n3 :|: n4 :|: n5 :|: n6 :|: n7 :|: n8 :|: n9 :|: N0
```
arianvp, line 280: So what if I have more than 9 endpoints? :)

serras: You can fall back to using a `NamedList`, so you would need to write `thing1 :|: .. :|: thing10 :|: N0`.

PR #174, opened 2020-04-07
Docs for GraphQL + reorganization

serras:


docs/docs/graphql.md
```diff
@@ -0,0 +1,158 @@
...
+  = Object "Hello" '[]
+           '[ Argument "who" Text :> Field "greeting" Text ]
+```
+
+This is expected: Mu's ability to target both RPC and GraphQL microservices means that sometimes there's some mismatch.
+
+[Morpheus GraphQL](https://morpheusgraphql.com/) also exposes GraphQL servers from Haskell code. Morpheus shared with Mu the ability to import a GraphQL schema into Haskell code. However, the types and fields are not represented by a type-level encoding, but *directly* as Haskell *records*.
+
+```haskell
+type Hello m = Hello { greeting :: Text -> m Text }
```
nalchevanidze, line 153: morpheus you need record Type for arguments.
```hs
data GreetingArgs = GreetingArgs{  argname :: Text } deriving (Generic, GQLType)
data Hello m = Hello { greeting :: GreetingArgs -> m Text } deriving (Generic, GQLType)
```

docs/docs/graphql.md
```diff
@@ -0,0 +1,158 @@
...
+
+```haskell
+type Hello
+  = Object "Hello" '[]
+           '[ Argument "who" Text :> Field "greeting" Text ]
+```
+
+This is expected: Mu's ability to target both RPC and GraphQL microservices means that sometimes there's some mismatch.
+
+[Morpheus GraphQL](https://morpheusgraphql.com/) also exposes GraphQL servers from Haskell code. Morpheus shared with Mu the ability to import a GraphQL schema into Haskell code. However, the types and fields are not represented by a type-level encoding, but *directly* as Haskell *records*.
```
nalchevanidze, line 150: additionally, morpheus has GraphQL Client and Subscriptions, but i don't know how relevant is it for `mu-haskell' context.

ps. you don't have to define schema in morpheus. it will be generated by  Breadth-first search on haskell types.

kutyel: we have subscriptions too 😉

docs/docs/grpc-client.md
```diff
@@ -1,24 +1,12 @@
...
-The combination of the declaration of a service API and a corresponding implementation as a `Server` may be served directly using a concrete wire protocol. One example is gRPC, provided by our sibling library `mu-grpc`. The following line starts a server at port `8080`, using Protocol Buffers as serialization layer:
-
-```haskell
-main = runGRpcApp msgProtoBuf 8080 quickstartServer
-```
-
-## Building a client
-
-Right now there are two options for building clients: using records or with `TypeApplications`. To give a proper introduction to both options let's consider in detail an example client for the following services:
+There are several options for building clients: you can choose between optics, records, and `TypeApplications`. To give a proper introduction to both options let's consider in detail an example client for the following services:
```
kutyel, line 9: On the second line, it mentions `both options` again 😅

docs/docs/grpc-server.md
```diff
@@ -0,0 +1,85 @@
...
+
+Being polymorphic here allows us to run the same server in multiple back-ends. Furthermore, by enlarging the set of abilities required for our monad `m`, we can [integrate with other libraries](transformer.md), including logging and resource pools.
+
+Since you can declare more than one method in a service, you need to join them into a `Server`. You do so by using `(:<|>:)` between each handler and ending the sequence with `H0`. In addition to the name of the service, `Server` has an additional parameter which records the types of the handlers. Since that list may become quite long, we can ask GHC to write it for us by using the `PartialTypeSignatures` extension and writing an underscore `_` in that position.
+
+```haskell
+{-# language PartialTypeSignatures #-}
+
+quickstartServer :: (MonadServer m) => ServerT QuickstartService m _
+quickstartServer = Server (sayHello :<|>: H0)
```
kutyel, line 33: This whole explanation about declaring services is outdated (if we merge your new PR), should I update it now or later? 😉

docs/docs/grpc-server.md
```diff
@@ -0,0 +1,85 @@
...
+sayHello :: (MonadServer m) => HelloRequest -> m HelloResponse
+sayHello (HelloRequest nm) = pure $ HelloResponse ("hi, " <> nm)
+```
+
+Notice the use of `MonadServer` in this case. This gives us the ability to:
+
+* Run arbitrary `IO` actions by using `liftIO`,
+* Return an error code by calling `serverError`.
+
+Being polymorphic here allows us to run the same server in multiple back-ends. Furthermore, by enlarging the set of abilities required for our monad `m`, we can [integrate with other libraries](transformer.md), including logging and resource pools.
```
kutyel, line 25: Will this link to `transformer.md` work right away?

docs/docs/grpc-server.md
```diff
@@ -0,0 +1,85 @@
...
+
+```haskell
+main = runGRpcApp msgProtoBuf 8080 quickstartServer
+```
+
+# Stream
+
+In the docs about [service definition]({% link docs/rpc.md %}) we had one single `SayHello` method which takes one value and produces one value. However, we can also declare methods which perform streaming, such as:
+
+```java
```
kutyel, line 49:
```suggestion
```protobuf
```

docs/docs/grpc-server.md
serras: My goal is to make this just the reorganization and GraphQL part. Then we merge and keep working on updating the rest.

docs/docs/grpc-server.md
serras: Nope, we need to use the special syntax.

serras:
This doesn't preclude you from using laziness, just use a type such as:

```haskell
data Author = Author AuthorId Text [Book+
```

However, combining laziness and `IO` is always a hard bussiness. Even more when this `IO` may need to read from the database or something like this. This design makes the programmer more aware of how and when the data is flowing.

docs/docs/graphql.md
serras: Thanks for the information! I've updated the docs :)

docs/docs/grpc-server.md
serras: Thanks for the catch!

docs/docs/grpc-server.md
serras: I've finally decided to include this into the docs.

PR #178, opened 2020-04-14
Use new server API in examples and upgrade deps ⚗️

kutyel:
Some examples like `seed` and `with-persistent` give a *ghcide* error on `import Mu.Schema.Optics`, this is an attempt to solve that and also a refactor of an orphan service with the old API.


kutyel:
@serras this needs some more work, since the ghcide error keeps appearing, also, some dependencies are set to 0.2, should we also update them all to 0.3? 🤔

serras:
Two comments:

1. Yes, we should update every dependency to 0.3.
2. I've thought about solving the error in `ghcide`, and the solution would involve a lot of refactoring. In particular, for each example project we would need to create a different folder for the `Schema.hs` file and a corresponding `library` stanza, and then one folder per executable. I really think that's too complicated for just having a couple fewer red lines in the code.

PR #182, opened 2020-04-17
Add new Stack template for GraphQL 🔥

kutyel:
🚀

### DIFF
--- a/templates/graphql-server.hsfiles
+++ b/templates/graphql-server.hsfiles
@@ -1,0 +1,94 @@
+{-# START_FILE {{name}}.cabal #-}
+name:               {{name}}
+version:            0.1.0.0
+
+-- synopsis:
+-- description:
+homepage:           https://github.com/{{github-username}}{{^github-username}}githubuser{{/github-username}}/{{name}}#readme
+author:             {{author-name}}{{^author-name}}Author name here{{/author-name}}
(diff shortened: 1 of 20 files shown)


templates/graphql-server.hsfiles
```diff
@@ -0,0 +1,94 @@
...
+
+main :: IO ()
+main = do
+  putStrLn "starting GraphQL server on port 8000"
+  run 8000 $ graphQLAppQuery server (Proxy @"Query")
+
+type ServiceMapping = '[]
+
+server :: forall m. MonadServer m => ServerT ServiceMapping ServiceDefinition m _
+server = resolver ( object @"Query" ( method @"hello" undefined ) )
```
serras, line 94: Better use `error "not implemented"` instead of `undefined`

templates/graphql-server.hsfiles
```diff
@@ -0,0 +1,94 @@
...
+import           Mu.Server
+
+graphql "MySchema" "ServiceDefinition" "schema.graphql"
+
+-- GraphQL App
+
+main :: IO ()
+main = do
+  putStrLn "starting GraphQL server on port 8000"
+  run 8000 $ graphQLAppQuery server (Proxy @"Query")
```
serras, line 89: There's a function `runGraphQLAppQuery` which makes you not require importing from Warp.

Issue #193, opened 2020-05-11
gRPC client does not send a message when calling a zero-arg Avro method


cb372:
## Scenario

Avro encoding.

Method definition (in `.avdl` format):

```
  void ping();
```

## Expected result

The HTTP/2 request should contain a DATA frame containing a 0-length gRPC message. So the DATA frame body should be 5 bytes long: 1 byte for compression flag + 4 bytes for the message length.

## Actual result

The request contains a 0-length DATA frame. There is no gRPC message, which is not valid behaviour.

Actually the [gRPC protocol spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) is a bit vague on this point, but the implication is there there will always be at least one message and never an empty DATA frame.

The server (Mu-Scala, which uses the official Java gRPC server implementation) certainly doesn't like it. It responds with status 13 (`INTERNAL`) and message `Half-closed without a request`.

serras:
This is indeed quite weird. The [implemention](https://github.com/haskell-grpc-native/http2-grpc-haskell/blob/master/http2-grpc-proto3-wire/src/Network/GRPC/HTTP2/Proto3Wire.hs#L48) of encoding always adds those two elements to the query.

serras:
Found it, there was some overriding of the default behavior especifically for empty tuple. https://github.com/higherkindness/mu-haskell/commit/821fcf5b396e91e98fa069f7b06be86bd78de1c6 should contain a fix.

cb372:
Fixed in #194

PR #194, opened 2020-05-19
Various fixes for Avro encoding

### LINKED ISSUE #192: Mysterious zero byte at start of Avro-encoded message
I'm upgrading the Mu-Scala integration tests to Mu-Haskell v0.3, and seeing some strange changes in behaviour since v0.2. Stand by for a few more issues this week 😄 

## Scenario

Mu-Scala server, Mu-Haskell client, Avro encoding.

## Request schema (in `.avdl` format)

```
  record GetForecastRequest {
    string city;
    int days_required;
  }
```

## Request message

```
GetForecastRequest(city = "London", days_required = 3)
```

## Expected Avro-encoded bytes

```
0c 4c 6f 6e 64 6f 6e 06
```

Explanation:

* `0c .. 6e` is the value of `city`, i.e. the string `London`
    * `0c` is the length, 6, in zig-zag encoding
* `06` is the value of `days_required`, i.e. `3` in zig-zag encoding

This is what Mu-Haskell v0.2 sends. It also matches what is generated by a Scala Avro library I tried.

## Actual bytes

```
00 0c 4c 6f 6e 64 6f 6e 06
```

Note the strange `00` byte at the start.

(Aside: the Mu-Haskell client compresses the message using GZip. So those 9 bytes get compressed (really poorly!) and end up as 29 bytes on the wire.)

## Result

Mu-Scala's Avro decoding implementation tries to decode these bytes as a list of messages, then use the head of the list as its result. So it decodes `00 0c` as `GetForecastRequest(city = "", days_required = 6)`, then blows up trying the decode the remaining bytes as another message.

## Packet captures

I've prepared Wireshark captures of Mu-Haskell v0.2 vs v0.3 but I can't attach them here. Let me know if you want them.

serras:
This is an attempt to fix #192 and #193

### DIFF
--- a/adapter/avro/mu-avro.cabal
+++ b/adapter/avro/mu-avro.cabal
@@ -42,6 +42,7 @@ library
     , template-haskell      >=2.12
     , text
     , time
+    , transformers
     , unordered-containers
     , uuid
     , vector
--- a/adapter/avro/src/Mu/Adapter/Avro.hs
+++ b/adapter/avro/src/Mu/Adapter/Avro.hs
@@ -21,7 +21,6 @@ and to Avro values.
 -}
 module Mu.Adapter.Avro () where
 
-import           Control.Applicative                 ((<|>))
 import           Control.Arrow                       ((***))
 import qualified Data.Avro                           as A
 import qualified Data.Avro.Encoding.FromAvro         as AVal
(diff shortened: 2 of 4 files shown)


cb372:
IIUC this PR is still missing the fix for decoding Avro messages, to stop it looking for a union discriminator as the first byte. When I ran the integration tests against f240916 I got the `Decoded Avro tag is outside the expected range for a Union` error I described in https://github.com/higherkindness/mu-haskell/issues/192#issuecomment-626834497

serras:
@kutyel @cb372 could you review the changes? The integration tests are passing now.

PR #200, opened 2020-06-17
Nix Strikes Back ❄️

kutyel:
I think we can give Nix a second chance and this way provide two usages to our users: **Nix and Stack** (since both will be tested on CI).

I'd like also to ping @domenkozar to see whether we can improve our Nix setup (maybe to avoid recompiling GHC everytime we update some deps) and also it would be marvelous to call [this python script](https://github.com/higherkindness/mu-haskell/blob/master/test-schema.sh) within Nix so it can be run on CI as well (right now it has just been ignored, I know @pepegar also told me how to run simple python scripts with Nix but I simply did not have more time to dig into this 😉 ).


pepegar:
Ah yep, what I meant is that, if we have nix-shell installed, a simple way of using that same script would be using nix-shell shebang, something like this maybe?

```
#! /usr/bin/env nix-shell
#! nix-shell -i python3 -p python3 python27Packages.avro python27Packages.protobuf
```

Or even translating it to a nix expression producing that shell script:

```nix
self: pkgs:
with pkgs;
let
  python3-packages = python-packages: with python-packages; [
    avro
  ];
  python3 = pkgs.python36Packages.python.withPackages python3-packages;
  python2-packages = python-packages: with python-packages; [
    protobuf
  ];
  python2 = pkgs.python27Packages.python.withPackages python2-packages;
  stack = pkgs.stack;
in
{
  test-schema = writeShellScriptBin "test-schema.sh" ''
  #!/bin/sh

  # In order to run this script you need
  # - avro for Python 3 https://avro.apache.org/releases.html
  # - protobuf for Python 2 https://github.com/protocolbuffers/protobuf/releases/
  #   follow https://github.com/protocolbuffers/protobuf/tree/master/python

  echo "BUILDING"
  ${stack}/bin/stack build mu-avro mu-protobuf
  mkdir -p dist

  echo "\nAVRO\n====\n"
  echo "python/generate"
  ${python3}/bin/python adapter/avro/test/avro/generate.py adapter/avro/test/avro/example.avsc dist/avro-python.avro
  ${stack}/bin/stack test-avro dist/avro-haskell.avro dist/avro-python.avro
  echo "ptyhon/consume"
  ${python3}/bin/python adapter/avro/test/avro/consume.py adapter/avro/test/avro/example.avsc dist/avro-haskell.avro

  echo "\nPROTOBUF\n========\n"
  echo "python/generate"
  ${python2}/bin/python adapter/protobuf/test/protobuf/generate.py dist/protobuf-python.pbuf
  ${stack}/bin/stack exec test-protobuf dist/protobuf-haskell.pbuf dist/protobuf-python.pbuf
  echo "python/consume"
  ${python2}/bin/python adapter/protobuf/test/protobuf/consume.py dist/protobuf-haskell.pbuf
  '';
}
```

I haven't tested neither of those... but ... that should work I believe

domenkozar:
For Python script, it's best described in https://nix.dev/tutorials/ad-hoc-developer-environments.html

As far as rebuilding GHC goes, it's best by pinning nixpkgs [as provided by haskell.nix](https://github.com/input-output-hk/haskell.nix#b-using-stackyaml) and configuring IOHK cache.

The alternative would be to pin nixpkgs only for building GHC, etc. but that's [yet to be done](https://github.com/input-output-hk/haskell.nix/issues/549)

PR #201, opened 2020-06-18
Add needed kind signatures

serras:
As part of the refactoring for 0.4, the type of names of arguments was generalized from being hardcoded to `Symbol` to being a parameter. This is required so that we can share the same data type as the type and term levels, since in the latter we need `Text` or `String` instead.

Unfortunately, this means that in some cases our quasi-quoters produce ambiguous types in those positions. Since they are supposed to always produce `Symbol`, I've added enough kind signatures to prevent the problem from happening.

Alas, now sometimes users have to enable the `KindSignatures` extension.


kutyel:
@serras please can we merge this after #200? that way we can test if the new CI works better 😄

PR #207, opened 2020-07-01
Add boundaries to all Cabal files 🚀

### LINKED ISSUE #191: Upload version boundaries to Hackage metadata for all `mu` packages
Use `cabal gen-bounds` to put PVP-compliant version bounds (lower
AND upper) on ALL your unique dependencies so your package will still
be buildable years down the road. One important thing to note is that
you only need to include version bounds once. For example, if you
depend on the same package in your library and your test suite, you
only need to put the version bounds for that dependency in one place.
This keeps the dependency bounds information DRY.

---

cc: @hvr 😉

kutyel:
Fixes #191 

## TODO*
- [x] `adapter/protobuf`
- [x] `grpc/client`
- [x] `grpc/common`
- [x] `grpc/server`

---
* Packages I wasn't able to generate boundaries automatically

### DIFF
--- a/adapter/avro/mu-avro.cabal
+++ b/adapter/avro/mu-avro.cabal
@@ -28,24 +28,24 @@ library
     Mu.Quasi.Avro.Example
 
   build-depends:
-      aeson
-    , avro                  >=0.5.1
+      aeson                 >=1.5.2   && <1.6
+    , avro                  >=0.5.2   && <0.6
     , base                  >=4.12    && <5
-    , bytestring
-    , containers
-    , deepseq
-    , language-avro         >=0.1.3.1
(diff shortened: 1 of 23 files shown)


serras:
In theory, the bounds over version 1.0 should be until the next major version. So you should have >= 1.8 && < 2.

How did you generate those initial bounds? Can we build the whole thing if we remove allow-newer?

Also, please, do not bump major versions of packages which did not change (like compendium-client).

kutyel:
I just run `cabal gen-bounds` on each package 😅

serras:
My main question is: is this still compatible with all the stack.yaml we have? I have the feeling that the lower bounds are a bit restrictive.

PR #210, opened 2020-07-16
Be more explicit about Protobuf defaults in docs and code

### LINKED ISSUE #209: Docs for auto-deriving 'Maybe a'
Hi, for some reason my custom `Maybe a` (where `a` is one of the primitive types - Text, ByteString, Int64) fields break auto-deriving of `ToSchema` and `FromSchema` protobuf bindings, and I couldn't find an example of such a type being used in examples listed in the official documentation.

Do I have to manually define the schema if I want a Maybe of a primitve type, as it's shown in https://github.com/higherkindness/mu-haskell/blob/b0f84869674868aa66c5387dbc1b74ca8840e4c8/core/schema/src/Mu/Schema/Examples.hs#L75 ?

serras:
Fixes #209 

pinging @coderfromhere

### DIFF
--- a/adapter/protobuf/src/Mu/Quasi/ProtoBuf.hs
+++ b/adapter/protobuf/src/Mu/Quasi/ProtoBuf.hs
@@ -18,6 +18,7 @@ module Mu.Quasi.ProtoBuf (
   , protobufToDecls
   ) where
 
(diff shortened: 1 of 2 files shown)


coderfromhere:
this looks great, thank you!

PR #211, opened 2020-07-24
Example using Persistent and GraphQL

### LINKED ISSUE #123: GraphQL: example applications
Take some examples from Apollo, Sangria, Caliban, or any other GraphQL functional framework, and turn it into a Mu+GraphQL application.

This issue is about the server part, the client part is covered by #86.

serras:
Fixes #123

### DIFF
--- a/examples/library/src/Schema.hs
+++ b/examples/library/src/Schema.hs
@@ -1,0 +1,53 @@
+{-# language CPP                        #-}
+{-# language DataKinds                  #-}
+{-# language DeriveAnyClass             #-}
+{-# language DeriveGeneric              #-}
+{-# language DerivingStrategies         #-}
+{-# language FlexibleContexts           #-}
+{-# language FlexibleInstances          #-}
+{-# language GADTs                      #-}
+{-# language GeneralizedNewtypeDeriving #-}
+{-# language MultiParamTypeClasses      #-}
+{-# language OverloadedStrings          #-}
+{-# language PartialTypeSignatures      #-}
+{-# language PolyKinds                  #-}
+{-# language QuasiQuotes                #-}
+{-# language ScopedTypeVariables        #-}
+{-# language StandaloneDeriving         #-}
+{-# language TemplateHaskell            #-}
+{-# language TupleSections              #-}
+{-# language TypeApplications           #-}
+{-# language TypeFamilies               #-}
+{-# language TypeOperators              #-}
+{-# language UndecidableInstances       #-}
+module Schema where
+
+import           Data.Int                (Int32, Int64)
+import qualified Data.Text               as T
+import           Database.Persist.Sqlite
+import           Database.Persist.TH
+import           GHC.Generics
+import           Mu.Adapter.Persistent   (WithEntityNestedId (..))
+import           Mu.GraphQL.Quasi
+import           Mu.Schema
+
+#if __GHCIDE__
+graphql "Library" "examples/library/library.graphql"
+#else
+graphql "Library" "library.graphql"
+#endif
+
+share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
+Author json
+  name T.Text
+  UniqueName name
+  deriving Show Generic
+Book json
+  title T.Text
+  author AuthorId
+  deriving Show Generic
+|]
+
+newtype NewAuthor = NewAuthor { name :: T.Text }
+  deriving stock (Eq, Show, Generic)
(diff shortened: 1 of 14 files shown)


examples/library/src/Schema.hs
```diff
@@ -0,0 +1,53 @@
...
+{-# language TypeOperators              #-}
+{-# language UndecidableInstances       #-}
+module Schema where
+
+import           Data.Int                (Int32, Int64)
+import qualified Data.Text               as T
+import           Database.Persist.Sqlite
+import           Database.Persist.TH
+import           GHC.Generics
+import           Mu.Adapter.Persistent   (WithEntityNestedId (..))
```
kutyel, line 30: Is this used?

examples/library/src/Schema.hs
```diff
@@ -0,0 +1,53 @@
...
+  UniqueName name
+  deriving Show Generic
+Book json
+  title T.Text
+  author AuthorId
+  deriving Show Generic
+|]
+
+newtype NewAuthor = NewAuthor { name :: T.Text }
+  deriving stock (Eq, Show, Generic)
```
kutyel, line 52: `deriving stock`? 😱

serras: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/deriving_strategies.html

examples/library/src/Server.hs
```diff
@@ -0,0 +1,107 @@
...
+
+type ObjectMapping = '[
+    "Book"   ':-> Entity Book
+  , "Author" ':-> Entity Author
+  ]
+
+libraryServer :: forall i.
+                 SqlBackend
+              -> ServerT ObjectMapping i Library ServerErrorIO _
+libraryServer conn
```
kutyel, line 56: Since now we have a complete example on a separate folder, can we remove the `Example.hs` in the graphql package?

examples/library/src/Server.hs
```diff
@@ -0,0 +1,107 @@
...
+    bookAuthor :: Entity Book -> ServerErrorIO (Entity Author)
+    bookAuthor (Entity _ Book { bookAuthor = a })
+      = runDb conn $ Entity a . fromJust <$> get a
+    authorId :: Entity Author -> ServerErrorIO Integer
+    authorId (Entity (AuthorKey k) _) = pure $ toInteger k
+    authorName :: Entity Author -> ServerErrorIO T.Text
+    authorName (Entity _ Author { authorName = t }) = pure t
+    authorBooks :: Entity Author -> ServerErrorIO [Entity Book]
+    authorBooks (Entity a _)
+      = runDb conn $ selectList [BookAuthor ==. a] [Asc BookTitle]
```
kutyel, line 82: Wow this is so nice and convenient! (I guess these are operators from `persistent`, right?) 🎉

examples/library/src/Server.hs
```diff
@@ -0,0 +1,107 @@
...
+      = runDb conn $ selectList [Filter AuthorName (FilterValue nameFilter) (BackendSpecificFilter "LIKE")]
+                                [Asc AuthorName]
+    allBooks :: T.Text -> ServerErrorIO [Entity Book]
+    allBooks titleFilter
+      = runDb conn $ selectList [Filter BookTitle (FilterValue titleFilter) (BackendSpecificFilter "LIKE")]
+                                [Asc BookTitle]
+    allBooksConduit :: ConduitT (Entity Book) Void ServerErrorIO () -> ServerErrorIO ()
+    allBooksConduit sink
+      = runDb conn $ runConduit $ selectSource [] [Asc BookTitle] .| transPipe raiseErrors sink
+      where raiseErrors :: forall m a. MonadIO m => ServerErrorIO a -> m a
```
kutyel, line 95: This looks very specific to `ServerErrorIO`, can we add it somewhere in the library maybe?

examples/library/src/Schema.hs
serras: Well spotted!

examples/library/src/Server.hs
serras: Yep, that's how you query stuff using `persistent`. Really magical 🧙

examples/library/src/Server.hs
serras: I've added it to the `mu-graphql` library. Unfortunately, one of the disadvantages we have when modelling streams is that the inner monad really *must* be `ServerErrorIO`, so lifting conduits is required.

PR #217, opened 2020-08-25
[WIP] Servant support

### LINKED ISSUE #215: Support generating RPC web servers from Mu Schemas
I want define my services and message types using .proto files (like with grpc) but use those endpoints from a browser (some combination of simple requests for unary RPCs and fetch/websockets for streaming RPCs). Does mu-haskell support this? I know there are products like grpc-web that allow you to communicate between a browser and a grpc server, but require a proxy to translate, but I would like to avoid using a proxy if possible. Now I don't know mu-haskell and its subcomponents very well, but I don't know of any reason why one couldn't convert proto->schema->http server (possibly utilizing the servant types?). Having this capability would greatly simplify the code I need to write for a web service I am working on, and could be immensely useful. If don't know of any product that makes proto services directly accessible from a browser since they all use grpc. If this is something that should be possible but simply hasn't been done I would be interested in diving deeper into mu-haskell to figure how to implement it. Thanks in advance.

andremarianiello:
I need to add support for putting paths in the API (probably something like /serviceName/methodName), but there is enough here to take a look. I would absolutely love it if the API could be generated from the service+schema only, but I didn't know how to convert a typeref into a haskell type, so I am currently using the handler types to direct that instead. Would appreciate any tips on that front. Additionally, we probably want to not restrict ourselves to just JSON, but I went with that for simplicity. Also, eventually, I would love to use protobuf options to direct the Service->API mapping, ala https://github.com/grpc-ecosystem/grpc-gateway, but I don't think that necessarily needs to be in this PR.

Fixes #215

### DIFF
--- a/core/rpc/mu-rpc.cabal
+++ b/core/rpc/mu-rpc.cabal
@@ -28,7 +28,8 @@ library
     Mu.Server
 
   build-depends:
-      base              >=4.12  && <5
+      aeson
+    , base              >=4.12  && <5
     , conduit           >=1.3.2 && <1.4
     , http-types        >=0.12  && <0.13
     , mtl               >=2.2   && <2.3
--- a/servant/server/mu-servant-server.cabal
+++ b/servant/server/mu-servant-server.cabal
@@ -47,6 +47,7 @@ executable servant-example-server
     , base
     , conduit
     , mu-rpc
+    , mu-schema
     , mu-servant-server
     , servant-server
     , text
--- a/servant/server/mu-servant-server.cabal
+++ b/servant/server/mu-servant-server.cabal
@@ -23,9 +23,11 @@ source-repository head
 library
   exposed-modules:  Mu.Servant.Server
   build-depends:
-      async
+      aeson
+    , async
     , base
     , conduit
+    , generic-aeson
     , mtl
     , mu-rpc
     , mu-schema
@@ -47,6 +49,7 @@ executable servant-example-server
     , mu-rpc
     , mu-servant-server
     , servant-server
+    , text
     , warp
 
   hs-source-dirs:   exe
--- a/servant/server/mu-servant-server.cabal
+++ b/servant/server/mu-servant-server.cabal
@@ -1,0 +1,54 @@
+name:               mu-servant-server
+version:            0.4.0.0
+synopsis:           Servant servers for Mu definitions
+description:
+  With @mu-servant-server@ you can easily build Servant servers for mu-haskell!
+
+license:            Apache-2.0
+license-file:       LICENSE
+author:             Temp Name
+maintainer:         alejandro.serrano@47deg.com
+copyright:          Copyright © 2019-2020 <http://47deg.com 47 Degrees>
+cabal-version:      >=1.10
+category:           Network
+build-type:         Simple
+extra-source-files: CHANGELOG.md
+homepage:           https://higherkindness.io/mu-haskell/
+bug-reports:        https://github.com/higherkindness/mu-haskell/issues
+
+source-repository head
+  type:     git
+  location: https://github.com/higherkindness/mu-haskell
+
+library
+  exposed-modules:  Mu.Servant.Server
+  build-depends:
+      async
+    , base
+    , conduit
+    , mtl
+    , mu-rpc
+    , mu-schema
+    , servant
+    , servant-server
+    , stm
+    , utf8-string
+
+  hs-source-dirs:   src
+  default-language: Haskell2010
+  ghc-options:      -Wall -fprint-potential-instances
+
+executable servant-example-server
+  main-is:          ExampleServer.hs
+  build-depends:
+      aeson
+    , base
+    , conduit
+    , mu-rpc
+    , mu-servant-server
+    , servant-server
+    , warp
+
+  hs-source-dirs:   exe
+  default-language: Haskell2010
+  ghc-options:      -Wall -fprint-explicit-kinds -fprint-explicit-foralls
--- a/servant/server/mu-servant-server.cabal
+++ b/servant/server/mu-servant-server.cabal
@@ -6,7 +6,7 @@ description:
 
 license:            Apache-2.0
 license-file:       LICENSE
-author:             Temp Name
+author:             Andre Marianiello
 maintainer:         alejandro.serrano@47deg.com
 copyright:          Copyright © 2019-2020 <http://47deg.com 47 Degrees>
 cabal-version:      >=1.10
--- a/servant/server/mu-servant-server.cabal
+++ b/servant/server/mu-servant-server.cabal
@@ -23,7 +23,7 @@ source-repository head
 library
   exposed-modules:  Mu.Servant.Server
   build-depends:
-      aeson
+      aeson                 >=1.4   && <2
     , async
     , base
     , conduit
--- a/servant/server/src/Mu/Servant/Server.hs
+++ b/servant/server/src/Mu/Servant/Server.hs
@@ -128,11 +128,12 @@ instance ServantMethodHandlers pkg svc m chn inh '[] '[] where
   servantMethodHandlers _ _ _ H0 = emptyServer
 
 instance
-  ( ServantMethodHandler httpMethod httpStatus m args ret h,
+  ( ServantMethodHandler httpMethod httpStatus ctype m args ret h,
     ServantMethodHandlers pkg sname m chn () rest hs,
     HttpMethodFor pkg sname mname ~ httpMethod,
     HttpStatusFor pkg sname mname ~ httpStatus,
-    Server (MethodAPI pkg sname ('Method mname args ret) h) ~ Server (HandlerAPI httpMethod httpStatus args ret h)
+    ContentTypeFor pkg sname mname ~ ctype,
+    Server (MethodAPI pkg sname ('Method mname args ret) h) ~ Server (HandlerAPI httpMethod httpStatus ctype args ret h)
   ) =>
   ServantMethodHandlers pkg sname m chn () ('Method mname args ret ': rest) (h ': hs)
   where
@@ -145,6 +146,7 @@ instance
       f
       (Proxy @httpMethod)
       (Proxy @httpStatus)
+      (Proxy @ctype)
       (Proxy @args)
       (Proxy @ret)
       (h NoRpcInfo ())
@@ -153,7 +155,14 @@ instance
 type family MethodAPI pkg sname method h where
   MethodAPI pkg sname ('Method mname args ret) h =
     PrefixRoute (RouteFor pkg sname mname)
-      (HandlerAPI (HttpMethodFor pkg sname mname) (HttpStatusFor pkg sname mname) args ret h)
+      ( HandlerAPI
+          (HttpMethodFor pkg sname mname)
+          (HttpStatusFor pkg sname mname)
+          (ContentTypeFor pkg sname mname)
+          args
+          ret
+          h
+      )
 
 type family HttpMethodFor pkg sname mname :: ServantMethod where
   HttpMethodFor pkg sname mname =
@@ -163,6 +172,19 @@ type family HttpStatusFor pkg sname mname :: ServantStatus where
   HttpStatusFor pkg sname mname =
     FromMaybe 200 (GetMethodAnnotationMay (AnnotatedPackage ServantStatus pkg) sname mname)
 
+type family ContentTypeFor pkg sname mname :: Type where
+  ContentTypeFor pkg sname mname =
+    UnwrapServantContentType
+      ( FromMaybe
+          ('ServantContentType JSON)
+          (GetMethodAnnotationMay (AnnotatedPackage ServantContentType pkg) sname mname)
+      )
+
+newtype ServantContentType = ServantContentType Type
+
+type family UnwrapServantContentType (sctype :: ServantContentType) :: Type where
+  UnwrapServantContentType ('ServantContentType ctype) = ctype
+
 type ServantMethod = StdMethod
 
 type ServantStatus = Nat
@@ -171,40 +193,49 @@ class
   ServantMethodHandler
     (httpMethod :: ServantMethod)
     (httpStatus :: ServantStatus)
+    (ctype :: Type)
     (m :: Type -> Type)
     (args :: [Argument snm anm tyref])
     (ret :: Return snm tyref)
     (h :: Type) where
-  type HandlerAPI (httpMethod :: ServantMethod) (httpStatus :: ServantStatus) args ret h
+  type
+    HandlerAPI
+      (httpMethod :: ServantMethod)
+      (httpStatus :: ServantStatus)
+      (ctype :: Type)
+      args
+      ret
+      h
   servantMethodHandler ::
     (forall a. m a -> Handler a) ->
     Proxy httpMethod ->
     Proxy httpStatus ->
+    Proxy ctype ->
     Proxy args ->
     Proxy ret ->
     h ->
-    Servant.Server (HandlerAPI httpMethod httpStatus args ret h)
+    Servant.Server (HandlerAPI httpMethod httpStatus ctype args ret h)
 
-instance ServantMethodHandler httpMethod httpStatus m '[] 'RetNothing (m ()) where
+instance ServantMethodHandler httpMethod httpStatus ctype m '[] 'RetNothing (m ()) where
   type
-    HandlerAPI httpMethod httpStatus '[] 'RetNothing (m ()) =
-      Verb httpMethod httpStatus '[JSON] ()
-  servantMethodHandler f _ _ _ _ = f
+    HandlerAPI httpMethod httpStatus ctype '[] 'RetNothing (m ()) =
+      Verb httpMethod httpStatus '[ctype] ()
+  servantMethodHandler f _ _ _ _ _ = f
 
-instance ServantMethodHandler httpMethod httpStatus m '[] ('RetSingle rref) (m r) where
+instance ServantMethodHandler httpMethod httpStatus ctype m '[] ('RetSingle rref) (m r) where
   type
-    HandlerAPI httpMethod httpStatus '[] ('RetSingle rref) (m r) =
+    HandlerAPI httpMethod httpStatus ctype '[] ('RetSingle rref) (m r) =
       Verb httpMethod httpStatus '[JSON] r
-  servantMethodHandler f _ _ _ _ = f
+  servantMethodHandler f _ _ _ _ _ = f
 
 instance
   (MonadServer m) =>
-  ServantMethodHandler httpMethod httpStatus m '[] ('RetStream rref) (ConduitT r Void m () -> m ())
+  ServantMethodHandler httpMethod httpStatus ctype m '[] ('RetStream rref) (ConduitT r Void m () -> m ())
   where
   type
-    HandlerAPI httpMethod httpStatus '[] ('RetStream rref) (ConduitT r Void m () -> m ()) =
+    HandlerAPI httpMethod httpStatus ctype '[] ('RetStream rref) (ConduitT r Void m () -> m ()) =
       Stream httpMethod httpStatus NewlineFraming JSON (SourceIO (StreamResult r))
-  servantMethodHandler f _ _ _ _ = liftIO . sinkToSource f
+  servantMethodHandler f _ _ _ _ _ = liftIO . sinkToSource f
 
 data StreamResult a = Error String | Result a
   deriving (Generic, Show)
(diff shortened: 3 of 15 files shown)


kutyel:
Hey @andremarianiello! Thanks for this great work, we'll review it, provide feedback and help you land this as soon as possible! 💪

andremarianiello:
Latest commit rewrote a lot of the stream handling. These changes were made to ensure that request body chunks are promptly delivered by the StreamBody SourceT, and that the StreamPost SourceT promptly delivered response body chunks. I was testing using the following command

```bash
strace -fy -e trace=network curl localhost:8081/streambody/streampost -XPOST -H 'content-type: application/json' --trace-ascii -H 'transfer-encoding: chunked' --no-buffer -v -sS -T - < request
```
where `request` is a fifo that I made with `mkfifo` and keep open for writing using `exec 3>request` in another shell. You can then send chunks of input with `echo '{"name": "me"}' >&3` to that fifo and see the response get printed by curl. If you don't put curl under strace then the response display lags by one request, but I wasn't able to figure out why. 

Additionally, I reworked the error handling a bit. Servant streams only have a HasServer instance for SourceIO. This is fine on the input SourceIO because we can raise the error as a Mu.Server.ServerError, but the output SourceIO is trickier. If we emit the error as a StepT Error then the response stream dies and the client has no idea why. Instead I opted for a technique that I have seen other places, which I implemented using `StreamResult`. Please let me know how you feel about that solution.

Finally, I added some dummy paths to the MethodAPI instances so that I could test the streaming - previously all the endpoint shared a handler and I could only run the first one. This is not a real solution though, so I plan to work on constructing a path from the rpc info next.

servant/server/mu-servant-server.cabal
```diff
@@ -0,0 +1,57 @@
+name:               mu-servant-server
+version:            0.4.0.0
+synopsis:           Servant servers for Mu definitions
+description:
+  With @mu-servant-server@ you can easily build Servant servers for mu-haskell!
+
+license:            Apache-2.0
+license-file:       LICENSE
+author:             Temp Name
```
kutyel, line 9: You can safely add your self as the author of this package 😉

serras:
First of all, this is awesome work! 💜 

There are two main issues I think should be ironed out:
1. As you mention, you use the handler types to guide conversion. However, it should never be the case that you need explicit Haskell types not `instance FromJSON HelloRequest` instances, so there's some work to be done there.
2. It's not clear to me how method names are mapped to routes. One possibility is to add some information as an annotation (as we do with GraphQL default values). Would that help in this case?

servant/server/mu-servant-server.cabal
andremarianiello: Haha this is the first time I'm authoring a package so I wasn''t sure. I'll update it.

andremarianiello:
> 2\. It's not clear to me how method names are mapped to routes. One possibility is to add some information as an annotation (as we do with GraphQL default values). Would that help in this case?

As a first pass, I think I am simply going to map them to /serviceName/methodName. However, a real user will definitely want control over not just the route, but all aspects of the Servant API type. I think it would make sense to provide a default mapping, but let the user override that and choose their own mapping, which I think annotations might work well for. Another possibility I was considering was letting users "tag" the handlers with the API type if they wanted. For example:

```haskell
sayHello :: Tagged ("ay" :> "hello" :> ReqBody HelloRequest :> Get '[JSON] HelloResponse) (HelloRequest -> m HelloResponse)
sayHello = ...
```
Typing it out i don't love the repetitive-ness. So I'll probably try the annotations approach first. Are there any plans to convert grpc options into annotations?

andremarianiello:
Some more changes. You can now choose a route on a per-rpc basis (see ExampleServer.hs for an example). However, the annotations are not required. If no service annotation is given, the service name is used as a route prefix. If no method annotation is given, the method name is used as a route suffix. I'm going to work on removing the handler types parameter next.

andremarianiello:
Pushed some more changes that make for a cleaner example server.

andremarianiello:
I pushed some changes to allow for customizing the http method or status of each rpc. Unfortunately, these require an AnnotatedPackage instance, even if its just `'[]`. I would be interested in hearing ideas on how to avoid that.

serras:
@andremarianiello does this build against `master` or against the 0.3 version?

andremarianiello:
@serras `master`

serras (commented):
The code looks great (I prefer to have `stylish-haskell` run over all files, but I can do it myself).

Before we merge I would love to rework some of the examples we have (health check, seed, library) to use this, as we did for exposing them via GraphQL. Could you provide a small explanation of what annotations do I need to write? I've seen [this code](https://github.com/higherkindness/mu-haskell/pull/217/files#diff-285ba3c9beda9ceb6306c29e592b467cR29) but it's not 100% clear what each element is.

Thanks!

andremarianiello:
Short version of this comment: all 5 instances in the ExampleServer are required, but the 3 AnnotatedPackage instances may be empty.

There are 3 annotation domains for packages and 2 for schemas. 

Package domains:
* ServantRoute
This is a type alias for `[Symbol]`, representing URL path components. An instance for this must be provided for your package (`AnnotatedPackage ServantRoute MyPackage`), but may be empty. It is fine to specify no annotations in the instance, but the instance is still required. The api for an rpc is located at the url path `/<serviceName>/<rpcName>` by default. Service annotations can alter the beginning of the path, or remove the service prefix altogether. Method annotations similarly change the end of the path.
* ServantMethod
This is a type alias for StdMethod from Servant, the type of HTTP methods (GET, POST, etc.). An instance for this domain must be provided, but may be empty. POST is the default method for all rpcs, but method annotations in the instance will override this. This can use used, for example, to use a GET on rpcs which you know to be read-only.
* ServantStatus
This is a type alias for Nat fromGHC.TypeLits, representing the return code on a successful response. An instance must be provided but may be empty. Method annotations in this instance override the default of 200.

Schema domains:
* ServantUnaryContentTypes
This holds a list of types representing valid content types (JSON, Plaintext, OctetStream) for a schema type when used as a unary HTTP request/response body. There is currently no default, so an annotation is required for each type in the schema that is used in a unary rpc request/response.
* ServantStreamContentType
This holds a type representing the content type to use when streaming this schema type, and it holds a type representing the streaming "framing", aka the way is which stream results are separated. There is no default, so an annotation is is required for each type that is used in a streaming rpc request/response.

servant/server/src/Mu/Servant/Server.hs
```diff
@@ -0,0 +1,351 @@
...
+{-# language PolyKinds             #-}
+{-# language RankNTypes            #-}
+{-# language ScopedTypeVariables   #-}
+{-# language TypeApplications      #-}
+{-# language TypeFamilies          #-}
+{-# language TypeOperators         #-}
+{-# language UndecidableInstances  #-}
+{-# OPTIONS_GHC -Wall #-}
+
+module Mu.Servant.Server where
```
serras, line 16: Could you list explicit exports?

serras, line 16: It would also be great if you could add the information in https://github.com/higherkindness/mu-haskell/pull/217#issuecomment-696128262 as part of the module documentation, so we do not forget.

stack.yaml
```diff
@@ -14,6 +14,7 @@ packages:
 - grpc/client
 - grpc/server
 - graphql
+- servant/server
```
serras, line 17: Please add this line also to `stack-14.yaml` and `stack-nightly.yaml`.

stack.yaml
```diff
@@ -37,6 +38,7 @@ extra-deps:
 - wai-middleware-prometheus-1.0.0
 - git: https://github.com/hasura/graphql-parser-hs.git
   commit: f4a093981ca5626982a17c2bfaad047cc0834a81
+- generic-aeson-0.2.0.10
```
serras, line 41: I didn't know about this library, but it's really nice! :)

core/rpc/mu-rpc.cabal
```diff
@@ -28,7 +28,8 @@ library
     Mu.Server
 
   build-depends:
-      base              >=4.12  && <5
+      aeson
```
serras, line 31: Please introduce bounds for this dependency. The best is to copy them from `mu-schema.cabal`.

servant/server/src/Mu/Servant/Server.hs
```diff
@@ -0,0 +1,351 @@
...
+{-# language GADTs                 #-}
+{-# language MultiParamTypeClasses #-}
+{-# language PolyKinds             #-}
+{-# language RankNTypes            #-}
+{-# language ScopedTypeVariables   #-}
+{-# language TypeApplications      #-}
+{-# language TypeFamilies          #-}
+{-# language TypeOperators         #-}
+{-# language UndecidableInstances  #-}
+{-# OPTIONS_GHC -Wall #-}
```
serras, line 14: We prefer to have this line as part of the `.cabal` file instead of per-file.

serras (changes requested):
Just a few details and this should be ready to merge 🚀

serras:
By the way, once this is merged we'll do a new major release of Mu. @andremarianiello do you have a Twitter 🐦  handle we could point to in order to acknowledge your awesome work in this?

servant/server/src/Mu/Servant/Server.hs
andremarianiello: Gladly. Ideally, I would like to use a custom `TypeError` to provide a nice error message when the user doesn't provide a certain instance. Currently, the error message looks like
```
/home/andre/mu-haskell/servant/server/exe/ExampleServer.hs:19:19: error:                                                                                                          • No instance for (HasServer                                                                                                                                                                       @{*}
                         (Mu.Servant.Server.PrefixRoute
                            (Mu.Servant.Server.Concat
                               @ghc-prim-0.5.3:GHC.Types.Symbol
                               (Mu.Servant.Server.FromMaybe
                                  @[ghc-prim-0.5.3:GHC.Types.Symbol]
                                  ((':)
                                     @ghc-prim-0.5.3:GHC.Types.Symbol
                                     "Greeter"
                                     ('[] @ghc-prim-0.5.3:GHC.Types.Symbol))
                                  (GetServiceAnnotationMay
                                     @ServantRoute
                                     @ghc-prim-0.5.3:GHC.Types.Symbol
                                     @ghc-prim-0.5.3:GHC.Types.Symbol
                                     @ghc-prim-0.5.3:GHC.Types.Symbol
                                     (AnnotatedPackage
                                        @ghc-prim-0.5.3:GHC.Types.Symbol
                                        @ghc-prim-0.5.3:GHC.Types.Symbol
                                        @ghc-prim-0.5.3:GHC.Types.Symbol
                                        @(Mu.Rpc.TypeRef ghc-prim-0.5.3:GHC.Types.Symbol)
                                        ServantRoute
                                        ('Mu.Rpc.Package
                                           @ghc-prim-0.5.3:GHC.Types.Symbol                                                                                                                                              @ghc-prim-0.5.3:GHC.Types.Symbol
                                           @ghc-prim-0.5.3:GHC.Types.Symbol                                                                                                                                              @(Mu.Rpc.TypeRef ghc-prim-0.5.3:GHC.Types.Symbol)
                                           ('Just @ghc-prim-0.5.3:GHC.Types.Symbol "helloworld")
                                           ((':)
                                              @(Mu.Rpc.Service
                                                  ghc-prim-0.5.3:GHC.Types.Symbol
                                                  ghc-prim-0.5.3:GHC.Types.Symbol
                                                  ghc-prim-0.5.3:GHC.Types.Symbol
                                                  (Mu.Rpc.TypeRef ghc-prim-0.5.3:GHC.Types.Symbol))
                                              ('Mu.Rpc.Service                                                                                                                                                                 @ghc-prim-0.5.3:GHC.Types.Symbol
                                                 @ghc-prim-0.5.3:GHC.Types.Symbol
                                                 @ghc-prim-0.5.3:GHC.Types.Symbol
                                                 @(Mu.Rpc.TypeRef ghc-prim-0.5.3:GHC.Types.Symbol)
                                                 "Greeter"
                                                 ((':)
                                                    @(Mu.Rpc.Method
                                                        @{*}
                                                        ghc-prim-0.5.3:GHC.Types.Symbol
                                                        ghc-prim-0.5.3:GHC.Types.Symbol
                                                        ghc-prim-0.5.3:GHC.Types.Symbol
                                                        (Mu.Rpc.TypeRef
                                                           ghc-prim-0.5.3:GHC.Types.Symbol))
                                                    ('Mu.Rpc.Method
                                                       @{*}
                                                       @ghc-prim-0.5.3:GHC.Types.Symbol
                                                       @ghc-prim-0.5.3:GHC.Types.Symbol
                                                       @ghc-prim-0.5.3:GHC.Types.Symbol
                                                       @(Mu.Rpc.TypeRef
                                                           ghc-prim-0.5.3:GHC.Types.Symbol)
                                                       "SayHello"
                                                       ((':)
                                                          @(Mu.Rpc.Argument
                                                              @{*}
                                                              ghc-prim-0.5.3:GHC.Types.Symbol
                                                              ghc-prim-0.5.3:GHC.Types.Symbol
                                                              (Mu.Rpc.TypeRef
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol))
                                                          ('Mu.Rpc.ArgSingle
                                                             @{*}
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             @(Mu.Rpc.TypeRef
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol)
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             ('Nothing
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol)
                                                             ('Mu.Rpc.SchemaRef
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                QuickstartSchema
                                                                "HelloRequest"))
                                                          ('[]
                                                             @(Mu.Rpc.Argument
                                                                 @{*}
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol
                                                                 (Mu.Rpc.TypeRef
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol))))
                                                       ('Mu.Rpc.RetSingle
                                                          @{*}
                                                          @(Mu.Rpc.TypeRef
                                                              ghc-prim-0.5.3:GHC.Types.Symbol)
                                                          @ghc-prim-0.5.3:GHC.Types.Symbol
                                                          ('Mu.Rpc.SchemaRef
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             QuickstartSchema
                                                             "HelloResponse")))
                                                    ((':)
                                                       @(Mu.Rpc.Method
                                                           @{*}
                                                           ghc-prim-0.5.3:GHC.Types.Symbol
                                                           ghc-prim-0.5.3:GHC.Types.Symbol
                                                           ghc-prim-0.5.3:GHC.Types.Symbol
                                                           (Mu.Rpc.TypeRef
                                                              ghc-prim-0.5.3:GHC.Types.Symbol))
                                                       ('Mu.Rpc.Method
                                                          @{*}
                                                          @ghc-prim-0.5.3:GHC.Types.Symbol
                                                          @ghc-prim-0.5.3:GHC.Types.Symbol
                                                          @ghc-prim-0.5.3:GHC.Types.Symbol
                                                          @(Mu.Rpc.TypeRef
                                                              ghc-prim-0.5.3:GHC.Types.Symbol)
                                                          "SayHi"
                                                          ((':)
                                                             @(Mu.Rpc.Argument
                                                                 @{*}
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol
                                                                 (Mu.Rpc.TypeRef
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol))
                                                             ('Mu.Rpc.ArgSingle
                                                                @{*}
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                @(Mu.Rpc.TypeRef
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol)
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                ('Nothing
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol)
                                                                ('Mu.Rpc.SchemaRef
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   QuickstartSchema
                                                                   "HiRequest"))
                                                             ('[]
                                                                @(Mu.Rpc.Argument
                                                                    @{*}
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol
                                                                    (Mu.Rpc.TypeRef
                                                                       ghc-prim-0.5.3:GHC.Types.Symbol))))
                                                          ('Mu.Rpc.RetStream
                                                             @{*}
                                                             @(Mu.Rpc.TypeRef
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol)
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             ('Mu.Rpc.SchemaRef
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                QuickstartSchema
                                                                "HelloResponse")))
                                                       ((':)
                                                          @(Mu.Rpc.Method
                                                              @{*}
                                                              ghc-prim-0.5.3:GHC.Types.Symbol
                                                              ghc-prim-0.5.3:GHC.Types.Symbol
                                                              ghc-prim-0.5.3:GHC.Types.Symbol
                                                              (Mu.Rpc.TypeRef
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol))
                                                          ('Mu.Rpc.Method
                                                             @{*}
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             @ghc-prim-0.5.3:GHC.Types.Symbol
                                                             @(Mu.Rpc.TypeRef
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol)
                                                             "SayManyHellos"
                                                             ((':)
                                                                @(Mu.Rpc.Argument
                                                                    @{*}
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol
                                                                    (Mu.Rpc.TypeRef
                                                                       ghc-prim-0.5.3:GHC.Types.Symbol))
                                                                ('Mu.Rpc.ArgStream
                                                                   @{*}
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   @(Mu.Rpc.TypeRef
                                                                       ghc-prim-0.5.3:GHC.Types.Symbol)
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   ('Nothing
                                                                      @ghc-prim-0.5.3:GHC.Types.Symbol)
                                                                   ('Mu.Rpc.SchemaRef
                                                                      @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                      @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                      @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                      QuickstartSchema
                                                                      "HelloRequest"))
                                                                ('[]
                                                                   @(Mu.Rpc.Argument
                                                                       @{*}
                                                                       ghc-prim-0.5.3:GHC.Types.Symbol
                                                                       ghc-prim-0.5.3:GHC.Types.Symbol
                                                                       (Mu.Rpc.TypeRef
                                                                          ghc-prim-0.5.3:GHC.Types.Symbol))))
                                                             ('Mu.Rpc.RetStream
                                                                @{*}
                                                                @(Mu.Rpc.TypeRef
                                                                    ghc-prim-0.5.3:GHC.Types.Symbol)
                                                                @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                ('Mu.Rpc.SchemaRef
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   @ghc-prim-0.5.3:GHC.Types.Symbol
                                                                   QuickstartSchema
                                                                   "HelloResponse")))
                                                          ('[]
                                                             @(Mu.Rpc.Method
                                                                 @{*}
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol
                                                                 ghc-prim-0.5.3:GHC.Types.Symbol
...
```

There is more but I got tired of copy-pasting it.  Point is, the error message the user gets when the AnnotatedPackage type family doesn't expand is completely unhelpful. Do you know any technique that would allow me to print a custom `TypeError` message when this happens? I don't know how to do that conditionally on a type family not expanding.

andremarianiello: Actually, I just found https://kcsongor.github.io/report-stuck-families/ which at first glance seems like it might have a solution.

serras: I was going to send you exactly that link, but you were faster!

andremarianiello: Pushed the custom error messages and documentation. I've never used haddock before, so please let me know if there is anything that I should do differently.

stack.yaml
andremarianiello: Done!

core/rpc/mu-rpc.cabal
andremarianiello: Done! Have you had issues with unbounded aeson versions in the past? I assumed stack would handle it.

servant/server/src/Mu/Servant/Server.hs
andremarianiello: Done!

andremarianiello:
> By the way, once this is merged we'll do a new major release of Mu. @andremarianiello do you have a Twitter 🐦 handle we could point to in order to acknowledge your awesome work in this?

Sorry, I am not on twitter. Best option I can offer is a link to my github profile page? I don't have time to do a lot of OSS contributions so I'm not sure if that is a reasonable solution, haha. Just my name with no link is fine as well, tbh. Thought I would appreciate it if you could let me know when you make the release announcement. I am looking forward to it!

core/rpc/mu-rpc.cabal
serras: You are 100% right. However, we are trying not to leave Cabal-only users out, and for those we want to follow the PVP and write bounds as much as we can.

kutyel:
Huge thanks @andremarianiello for your massive work here! 🙌  we owe you a 🍻  if you come to the next Lambda World! (or any other event that 47 Degrees organises 😂 ) 👏 👏 👏

PR #220, opened 2020-09-04
Update instructions about dev environment

serras:


kutyel (approved):
I was thinking about updating this a while ago, I'm glad you found time to do that! 🚀

PR #222, opened 2020-09-04
Keep all annotations from .proto files

### LINKED ISSUE #216: Make gRPC options defined in proto files available
It would be nice of the grpc parser didn't throw away information stored in the grpc options. When working with proto files in other languages, field options, rpc options, etc. have been incredibly useful for adding things like validation and access control onto protobuf types and grpc endpoints. Custom options for protobufs are one of its most powerful features, and it would be a shame if they couldn't be leveraged.

serras:
Fixes #216


kutyel (approved):
:shipit:

PR #223, opened 2020-09-04
Conversion to logical types

### LINKED ISSUE #208: Docs for mapping custom types to/from primitive protocol types
Hi,  the current docstring sections for `ToSchema` and `FromSchema` don't explain the way I can define a conversion to/from protobuf `bytes` and Haskell `Data.UUID.UUID`. How would I define such a type? How would I reject incorrect byte values that are not valid UUID?

serras:
Fixes #208

This PR adds a new possibility when converting Haskell types from and to primitive types. Right now if the protocol defines that a certain type (let's say `bytes` in Protocol Buffers) maps to a certain Haskell types (`ByteString` in this case), the Haskell representation is forced to used that same type or write the conversion entirely by hand. When this is merged, you'll be able to do:

```haskell
data Msg = Msg { id :: Underlying ByteString UUID }
```

which specifies that on the wire we want to see the field as a `ByteString` but logically it is a `UUID`.

The conversion between wire and logical types is done via the `UnderlyingConversion` type class. Right now there are only instances for `UUID`, but more could be added here.

### DIFF
--- a/core/schema/mu-schema.cabal
+++ b/core/schema/mu-schema.cabal
@@ -1,5 +1,5 @@
 name:               mu-schema
-version:            0.3.0.0
+version:            0.3.1.0
 synopsis:           Format-independent schemas for serialization
 description:
   With @mu-schema@ you can describe schemas using type-level constructs, and derive serializers from those. See @mu-avro@, @mu-protobuf@ for the actual adapters.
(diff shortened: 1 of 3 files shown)


kutyel (approved):
🔥

PR #227, opened 2020-09-17
Mu.Schema.Lens package

andremarianiello:
I made this in an attempt to provide an equivalent to Mu.Schema.Optics, but for people who use `lens`. I will understand if you don't want it, but if you don't please let me know so that I can make it available some other way. Thanks!

### DIFF
--- a/examples/seed/src/Lens.hs
+++ b/examples/seed/src/Lens.hs
@@ -1,0 +1,77 @@
+{-# language DataKinds             #-}
+{-# language DuplicateRecordFields #-}
+{-# language FlexibleContexts      #-}
+{-# language OverloadedLabels      #-}
+{-# language OverloadedStrings     #-}
+{-# language PartialTypeSignatures #-}
+{-# language TypeApplications      #-}
+{-# language TypeOperators         #-}
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+
+module Main where
+
+import           Control.Concurrent       (threadDelay)
+import           Control.Lens
+import           Control.Monad.IO.Class   (liftIO)
+import           Control.Monad.Logger
+import           Data.Conduit
+import           Data.Conduit.Combinators as C
+import           Data.Text                as T
+import           Mu.GRpc.Server
+import           Mu.Schema
+import           Mu.Schema.Lens
+import           Mu.Server
+
+import           Schema
+
+type Person          = Term SeedSchema (SeedSchema :/: "Person")
+type PeopleRequest   = Term SeedSchema (SeedSchema :/: "PeopleRequest")
+type PeopleResponse  = Term SeedSchema (SeedSchema :/: "PeopleResponse")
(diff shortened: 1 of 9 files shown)


serras:
The code looks OK, as far as I understand what's going on. However, when using it I get ambiguity errors all over the place. As an example, I've ported the `seed-server` applications to lens in the [`mu-lens` branch](https://github.com/higherkindness/mu-haskell/tree/mu-lens), assuming that `mu-lens` is to be used as `mu-optics`. Unfortunately I get about 5 of these:

```
/Users/serras/Projects/mu-haskell/examples/seed/src/Lens.hs:71:11: error:
    • Couldn't match type ‘Uninterpret r0’ with ‘'TPrimitive Text’
        arising from a use of ‘record’
      The type variable ‘r0’ is ambiguous
    • In the second argument of ‘($)’, namely ‘record (go w)’
      In the expression: pure $ record (go w)
      In an equation for ‘getWeather’:
          getWeather msg
            | Just w <- msg ^. #currentWeather = pure $ record (go w)
            | otherwise = pure $ record "who knows?"
            where ...
   |
71 |  = pure $ record (go w)
   |           ^^^^^^^^^^^^^
```

I might be doing something wrong by assuming that I can just copy the code that uses `mu-optics` unchanged, though.

serras (changes requested):
I wrote some comments. In addition, would it be possible for you to run `stylish-haskell` with our custom set of rules in the `Lens.hs` file?

andremarianiello:
@serras I will take a look at those ambiguity errors to see if I can fix them. I'll handle the stylish-haskell formatting as well. Thanks for the feedback!

andremarianiello:
@serras One set of issues was handled by adding missing fundeps on the BuildRecord class, and one set was handled by adding a lens version of `is` from mu-optics. I cherry-picked your commit into my branch and included the fixes, and the example builds now.                                  Do you want me to add a version of the persistent-client-optics example as well?

serras:
@andremarianiello thanks for all your hard work! There's no need to update the other example (although we would not complain if you do, of course!). I'll review the code a bit more tomorrow, but I think it's ready to be merged :)

examples/seed/src/Lens.hs
```diff
@@ -71,7 +71,7 @@ getWeather msg
  = pure $ record (go w)
  | otherwise
  = pure $ record "who knows?"
- where go e | not $ e `isn't` #sunny  = "is sunny! 😄"
-            | not $ e `isn't` #cloudy = "is cloudy 😟"
-            | not $ e `isn't` #rainy  = "is rainy... 😭"
+ where go e | e `is` #sunny  = "is sunny! 😄"
```
kutyel, line 74: I think it looks way better this way! 🎉

serras (changes requested):
The last tiny bit: could you use a explicit export for `Mu.Schema.Lens`?

serras (approved):
Awesome!

PR #228, opened 2020-09-21
Fix and update stack.yaml and nix

serras:


kutyel:
I guess this works locally? 😅

serras:
It turns out that I had to change the components because “all” seems not to be available any more. Now it builds.

kutyel (approved):
:shipit:

PR #229, opened 2020-09-25
Port health check to use Servant + fix a few corner cases

serras:
You might be interested in this @andremarianiello 😉 

I added REST support to the the `health-check` example using the new Servant support. I can only say: it's awesome so easy it is, just specify routes and verbs and it works!

Well... I had to fix a couple of corner cases in the code, but that's how we make progress.

### DIFF
--- a/examples/health-check/mu-example-health-check.cabal
+++ b/examples/health-check/mu-example-health-check.cabal
@@ -21,25 +21,28 @@ executable health-server
   main-is:          Server.hs
   other-modules:    Definition
   build-depends:
-      async           >=2.2   && <3
-    , base            >=4.12  && <5
-    , conduit         >=1.3.2 && <2
-    , deferred-folds  >=0.9   && <0.10
-    , mu-graphql      >=0.4.0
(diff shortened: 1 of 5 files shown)


kutyel (approved):
:shipit:

andremarianiello:
Looks sweet!

andremarianiello:
Looking at what you had to do for a more real example, I wonder if the ServantRoute, ServantMethod, and ServantStatus types should all be combined into one. Having the path separate from the verb is harder to read (and more work to write) than just simply
```
'AnnMethod "MyService" "MyMethod" ('ServantRPCSettings '["my_method"] 'GET 200])
```
though maybe that is just me. I don't think it should necessarily be changed at this point, but I would be surprised if users complain about it. Though if they do, I bet someone will have good suggestions on how to improve the UX. Maybe its best to wait and see.

serras:
I think that's a very good idea, I also felt that just 2 type families should be enough. Let me try to produce a patch and then @andremarianiello @kutyel you tell me what you think.

PR #230, opened 2020-09-28
Simpler routes in Servant

serras:
As discussed in #229, the information for serving a Mu server via Servant could be condensed in two annotations: one for the routes, and one for the serialization of schemas. This PR implements it, and also introduces a `DefaultServantContentTypes` for those who don't want to think about which content types to support (simply JSON on both unary and streaming mode).

Pinging @andremarianiello for comments.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -33,5 +33,6 @@ in {
   mu-protobuf = hnPkgs.mu-protobuf.components.library;
   mu-rpc = hnPkgs.mu-rpc.components.library;
   mu-schema = hnPkgs.mu-schema.components.library;
+  mu-servant-server = hnPkgs.mu-servant-server.components.library;
   mu-tracing = hnPkgs.mu-tracing.components.library;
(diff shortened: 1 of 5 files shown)


serras:
As an example, the `health-check` server now looks like this https://github.com/higherkindness/mu-haskell/pull/230/files#diff-4d82008868163185644a0cbb894abcd8R154

kutyel (dismissed):
Looks good, could you also add `mu-servant` to our `default.nix`? 👍

serras:
> Looks good, could you also add `mu-servant` to our `default.nix`? 👍

Done! Let's wait to see whether Nix accepts this package too.

kutyel (approved):
:shipit:

andremarianiello:
I think this makes the example much nicer to write and read. Sorry I couldn't comment before the merge, I think our timezones are way off, haha

PR #231, opened 2020-09-28
Expose Swagger definition + SwaggerUI example

serras:
It turns out that turning @andremarianiello's work into a full OpenAPI server was quite easy, thanks to `servant-swagger`. This adds the necessary boilerplate in `mu-servant-server`, and also shows how to use `servant-swaggerui` to expose the nice UI from Swagger.

### DIFF
--- a/examples/health-check/src/Definition.hs
+++ b/examples/health-check/src/Definition.hs
@@ -15,6 +15,7 @@
 module Definition where
 
 import qualified Data.Aeson      as J
+import qualified Data.Swagger    as Swagger
 import           Data.Text       as T
 import           GHC.Generics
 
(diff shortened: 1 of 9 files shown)


serras:
Marking it as WIP since I want to add docs

examples/health-check/src/Definition.hs
```diff
@@ -34,27 +35,31 @@ newtype HealthCheckMsg
   = HealthCheckMsg { nameService :: T.Text }
   deriving ( Eq, Show, Ord, Generic
            , ToSchema   HealthCheckSchema "HealthCheck"
-           , FromSchema HealthCheckSchema "HealthCheck" )
+           , FromSchema HealthCheckSchema "HealthCheck"
+           , Swagger.ToSchema )
```
kutyel, line 39: hahahaha amazing, just by adding `deriving Swagger.ToSchema` we get all this magic! 🚀

examples/health-check/src/Server.hs
```diff
@@ -151,7 +154,8 @@ instance MonadMonitor m => MonadMonitor (TraceT m)
 -- Information for servant
 
 type instance AnnotatedPackage ServantRoute HealthCheckServiceFS2
-  = '[ 'AnnService "HealthCheckServiceFS2"
+  = '[ 'AnnPackage ('ServantAdditional ('Just (SwaggerSchemaUI "swagger-ui" "swagger.json")))
```
kutyel, line 157: Wow more type-level!

serras: Yes, that's how you add additional routes to the server which are not present in the API itself.

servant/server/src/Mu/Servant/Server.hs
```diff
@@ -445,7 +478,8 @@ sourceToSource (SourceT src) = ConduitT (PipeM (liftIO $ src (pure . go)) >>=)
 -- 2. HTTP method which must be used,
 -- 3. HTTP status code of a successful HTTP response from a specific `Method`. Use 200 for the usual status code.
 data ServantRoute
-  = ServantTopLevelRoute [Symbol]
+  = ServantAdditional (Maybe Type)
```
andremarianiello, line 481: Instead of using `Maybe Type` with `'Nothing` as the identity API, what about just using `Type` with `EmptyAPI` as the identity?

serras: Great idea! I've implemented it in the last commit.

kutyel (approved):
🔥

serras:
@andremarianiello as part of this PR I turned your comments in the module into documentation for our website. If you are curious, you can see the result [here](https://higherkindness.io/mu-haskell/wip/openapi/).

PR #233, opened 2020-10-12
Prepare release 0.4

serras:


serras:
@kutyel you might be happy to know that we now have a [`release-package` script](https://github.com/higherkindness/mu-haskell/blob/master/release-package.sh), so now releasing a new version of one package is as simple as doing `./release-package.sh mu-schema 0.3.1.0`.

PR #236, opened 2020-10-17
Improve the Library example (with persistent)

VladimirLogachev:
Hi!

My goal is to have a comprehensive example to use with different clients (such as with `elm-graphql` and other languages). So I changed this:

- Add `newBook` mutation
- Insert demo data to be instantly available when server starts (same demo data as in `library-graphql` executable) 
- Add uniqueness constraint for a tuple of a book title and book author id

### DIFF
--- a/examples/library/src/Server.hs
+++ b/examples/library/src/Server.hs
@@ -10,7 +10,6 @@
 
 module Main where
 
-import           Control.Monad.Except              (catchError)
 import           Control.Monad.IO.Class            (liftIO)
 import           Control.Monad.Logger
 import           Data.Conduit
@@ -69,9 +68,11 @@ Returns Nothing in case of any failure, including attempts to insert non-unique 
 -}
 insertAuthorAndBooks :: SqlBackend -> Author -> [Key Author -> Book] -> LoggingT IO (Maybe ())
 insertAuthorAndBooks conn author books =
-  (`catchError` (const $ pure Nothing)) . runDb conn . fmap sequence_ $ do
-    Just authorId <- insertUnique author
-    traverse (insertUnique . ($ authorId)) books
+  runDb conn . fmap sequence_ $ do
+    authorResult <- insertUnique author
+    case authorResult of
+      Just authorId -> traverse (insertUnique . ($ authorId)) books
+      Nothing       -> pure [Nothing]
 
 
 type ObjectMapping = '[
--- a/examples/library/src/Server.hs
+++ b/examples/library/src/Server.hs
@@ -10,6 +10,7 @@
 
 module Main where
 
+import           Control.Monad.Except              (catchError)
 import           Control.Monad.IO.Class            (liftIO)
 import           Control.Monad.Logger
 import           Data.Conduit
@@ -30,7 +31,6 @@ import           Schema
 
 main :: IO ()
 main = do
-  putStrLn "starting GraphQL server on port 8000"
   -- Setup CORS
   let hm = addHeaders [
              ("Access-Control-Allow-Origin", "*")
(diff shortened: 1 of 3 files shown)


examples/library/src/Server.hs
```diff
@@ -42,12 +42,37 @@ main = do
...
+  , insertAuthorAndBooks conn (Author "Michael Ende")
+      [Book "The Neverending Story", Book "Momo"]
+  ]
+
+{- | Inserts Author and Books
+Returns Nothing in case of any failure, including attempts to insert non-unique values
+-}
+insertAuthorAndBooks :: SqlBackend -> Author -> [Key Author -> Book] -> LoggingT IO (Maybe ())
+insertAuthorAndBooks conn author books =
+  (`catchError` (const $ return Nothing)) . runDb conn . fmap sequence_ $ do
```
kutyel, line 72:
```suggestion
  (`catchError` (const $ pure Nothing)) . runDb conn . fmap sequence_ $ do
```

examples/library/src/Server.hs
```diff
@@ -42,12 +42,37 @@ main = do
...
+    Just authorId <- insertUnique author
+    mapM (insertUnique . ($ authorId)) books
```
kutyel, line 74:
```suggestion
    traverse (insertUnique . ($ authorId)) books
```

kutyel (changes requested):
Looks great! 👏  I only added a few minor stylistic changes to match the rest of the mu-haskell codebase 😉

kutyel (dismissed):
🚀  @serras what do you think? :)

examples/library/src/Server.hs
```diff
@@ -42,12 +40,36 @@ main = do
...
+      [Book "Critique of Pure Reason"]
+  , insertAuthorAndBooks conn (Author "Michael Ende")
+      [Book "The Neverending Story", Book "Momo"]
+  ]
+
+{- | Inserts Author and Books
+Returns Nothing in case of any failure, including attempts to insert non-unique values
+-}
+insertAuthorAndBooks :: SqlBackend -> Author -> [Key Author -> Book] -> LoggingT IO (Maybe ())
+insertAuthorAndBooks conn author books =
```
serras, line 69: Would it be possible to re-write this function in a simpler way? Maybe using pattern matching on `authorResult`, as it was at some point?

Our goal with our examples is to showcase the library, but the readers may not be Haskell experts. Thus we try to keep the level of the language as low as possible.

examples/library/src/Server.hs
```diff
@@ -42,12 +40,36 @@ main = do
...
+  runDb conn $ (>>= sequence_) <$> do
+    authorResult <- insertUnique author
+    traverse (\authorId -> traverse (insertUnique . ($ authorId)) books) authorResult
```
serras, line 72: Similar here, we could write a much easier version by eta-expanding.

```haskell
   insertUnique . ($ authorId)
= \kBook -> insertUnique (kBook $ authorId)
= \kBook -> insertUnique (kBook authorId)
```

serras (changes requested):

examples/library/src/Server.hs
VladimirLogachev: Sure!

examples/library/src/Server.hs
VladimirLogachev: Changed it, thank you.

kutyel (approved):

serras (approved):

serras:
Thanks for your help! I'll merge once CI passes.

VladimirLogachev:
Is there an issue with CI? Maybe it is not going to run on forks?

VladimirLogachev:
If it's true that CI pipeline isn't going to run automatically on forks, and also can not be run manually, then maybe we may have a special branch for that? Call it `incoming`, `ci`, `integration` (or like that) for one specific reason: run CI pipeline before merging into master. Then such a branch could become a target branch for this PR.

kutyel:
Or we can just merge it and see what happens 😂  thanks for your work!

PR #246, opened 2020-11-08
protobuf: Translate unsigned integers to corresponding Word types

### LINKED ISSUE #245: Support unsigned integers in protobuf?
I see that unsigned integers are not supported in protobuf and it [looks intentional](https://github.com/higherkindness/mu-haskell/blob/678cd71f5c0628c05d77298fb9e91155801dfd1a/adapter/protobuf/src/Mu/Quasi/ProtoBuf.hs#L115).
I think (perhaps naively) that they can simply be represented by `Word32` and `Word64` from `Data.Word`. Is there a reason why this doesn't work?

akshaymankar:
Fixes #245


serras (approved):
LGTM

PR #248, opened 2020-11-08
library example: Use serverError instead of optional results

VladimirLogachev:
Hi!

I consider (using mandatory results for mutations + providing error messages) to be a better practice than returning nullable results because it explains the reason for failure.
Also, It's a bit easier to implement clients when results are required because in the case of `Nothing` the `newBook` or `newAuthor` field completely disappears from the JSON `data` object (though I understand that this case is both an encoding nuance and a decoding issue).


kutyel (approved):
LGTM, what do you think @serras? 😉

serras:
Yes, those are conditions which really should be satisfied, so it makes sense to error out.

serras:
@VladimirLogachev I just wanted to reiterate how grateful we are for the time you are taking to improve Mu-Haskell 😸

PR #249, opened 2020-11-09
Improve GraphQL example library 📚

kutyel:
Improve a little the code and being more explicits about the importing libraries 😄

### DIFF
--- a/examples/library/src/Server.hs
+++ b/examples/library/src/Server.hs
@@ -20,10 +20,7 @@ import           Mu.Adapter.Persistent             (runDb)
 import           Mu.GraphQL.Server                 (graphQLApp, liftServerConduit)
 import           Mu.Instrumentation.Prometheus     (initPrometheus, prometheus)
 import           Mu.Schema                         (Mapping ((:->)), Proxy (Proxy))
-import           Mu.Server                         (ServerError (ServerError),
-                                                    ServerErrorCode (Invalid), ServerErrorIO,
-                                                    ServerT, field, method, object, resolver,
-                                                    serverError)
+import           Mu.Server
 import           Network.Wai.Handler.Warp          (run)
 import           Network.Wai.Middleware.AddHeaders (addHeaders)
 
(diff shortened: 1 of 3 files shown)


examples/library/src/Server.hs
```diff
@@ -11,18 +11,19 @@
...
 import           Mu.Adapter.Persistent             (runDb)
-import           Mu.GraphQL.Quasi
-import           Mu.GraphQL.Server
-import           Mu.Instrumentation.Prometheus
-import           Mu.Schema
-import           Mu.Server
+import           Mu.GraphQL.Server                 (graphQLApp, liftServerConduit)
+import           Mu.Instrumentation.Prometheus     (initPrometheus, prometheus)
+import           Mu.Schema                         (Mapping ((:->)), Proxy (Proxy))
+import           Mu.Server                         (ServerError (ServerError),
```
serras, line 23: If any, I would not use explicit exports for the `Mu` things, the list is too long and we use almost everything.

serras (approved):

examples/library/src/Server.hs
```diff
@@ -1,12 +1,9 @@
 {-# language DataKinds             #-}
-{-# language FlexibleContexts      #-}
 {-# language OverloadedStrings     #-}
 {-# language PartialTypeSignatures #-}
-{-# language ScopedTypeVariables   #-}
 {-# language TypeApplications      #-}
-{-# language TypeFamilies          #-}
 {-# language TypeOperators         #-}
-{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+{-# options_ghc -fno-warn-partial-type-signatures #-}
```
serras, line 6: Can you actually do this? I thought that had to be in uppercase.

serras (approved):

examples/library/src/Server.hs
kutyel: It works 🤷🏻‍♂️  😆

PR #250, opened 2020-11-10
📦💥  Defend against possible explosion of decodeUtf8

kutyel:
I read here https://www.snoyman.com/blog/2020/11/haskell-bad-parts-2 that to my surprise `decodeUtf8` could blow up, and found that we use it in `mu-graphql`, so I thought it would be a good idea to fix it before it explodes 😉 

Thanks @snoyberg!


serras (approved):
Please check that the examples keep working before merging. Otherwise, LGTM.

PR #252, opened 2020-11-12
📚 📸  Add book images to graphql library example 

kutyel:
Let's make our example even more beautiful!

### DIFF
--- a/graphql/src/Mu/GraphQL/Server.hs
+++ b/graphql/src/Mu/GraphQL/Server.hs
@@ -45,10 +45,10 @@ import qualified Data.HashMap.Strict              as HM
 import           Data.Proxy                       (Proxy (..))
 import qualified Data.Text                        as T
 import           Data.Text.Encoding               (decodeUtf8')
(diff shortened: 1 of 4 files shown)


graphql/src/Mu/GraphQL/Server.hs
```diff
@@ -45,10 +45,10 @@ import qualified Data.HashMap.Strict              as HM
 import           Data.Proxy                       (Proxy (..))
 import qualified Data.Text                        as T
 import           Data.Text.Encoding               (decodeUtf8')
+import           Data.Text.Encoding.Error         (UnicodeException (..))
```
serras, line 48: why do you need this change in this PR?

serras (approved):

graphql/src/Mu/GraphQL/Server.hs
kutyel: It was a leftover from my previous PR, my OCD told me that the imports needed to be alphabetised 🤪

PR #253, opened 2020-11-12
Loosen bounds, update dependencies

serras:
As part of my on-going investigation of WebSockets, I've been updating the dependencies, and loosening some bounds in preparation for new versions of libraries in the next LTS.


serras:
The goal was to minimize the amount of `allow-newer: true` we are using. I'm also sending PRs to different libraries to loosen their bounds and try to get them in Stackage.

PR #256, opened 2020-11-12
Use a published GraphQL parser

serras:
Finally we can drop the dependency on a particular commit of a repo!

### DIFF
--- a/graphql/src/Mu/GraphQL/Annotations.hs
+++ b/graphql/src/Mu/GraphQL/Annotations.hs
@@ -5,7 +5,6 @@
 {-# language TupleSections       #-}
 {-# language TypeApplications    #-}
 {-# language TypeOperators       #-}
-{-# language ViewPatterns        #-}
 {-|
 Description : Annotations for GraphQL services
 
@@ -23,12 +22,11 @@ module Mu.GraphQL.Annotations (
 , module Mu.Rpc.Annotations
 ) where
 
-import           Control.Applicative           (Alternative (..))
-import           Data.Coerce
+import           Control.Applicative  (Alternative (..))
 import           Data.Proxy
-import qualified Data.Text                     as T
+import qualified Data.Text            as T
 import           GHC.TypeLits
-import qualified Language.GraphQL.Draft.Syntax as GQL
+import qualified Language.GraphQL.AST as GQL
 
 import           Mu.Rpc.Annotations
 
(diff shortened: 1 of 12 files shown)


graphql/src/Mu/GraphQL/Annotations.hs
```diff
@@ -110,11 +108,14 @@ instance (ReflectValueConst x, ReflectValueConstList xs)
...
   reflectValueConstObject _ = []
 instance (KnownSymbol a, ReflectValueConst x, ReflectValueConstObject xs)
          => ReflectValueConstObject ( '(a, x) ': xs) where
   reflectValueConstObject _
-    = GQL.ObjectFieldG (coerce $ T.pack $ symbolVal (Proxy @a)) (reflectValueConst (Proxy @x))
+    = GQL.ObjectField (T.pack $ symbolVal (Proxy @a))
+                      (GQL.Node (reflectValueConst (Proxy @x)) zl)
+                      zl
       : reflectValueConstObject (Proxy @xs)
+    where zl = GQL.Location 0 0
```
kutyel, line 121: I don't understand this bit 🤔

graphql/src/Mu/GraphQL/Server.hs
```diff
@@ -175,6 +175,7 @@ httpGraphQLAppTrans f server q m s req res =
     toResponse = res . responseBuilder ok200 [] . T.encodeUtf8Builder . encodeToLazyText
     unpackUnicodeException :: UnicodeException -> T.Text
     unpackUnicodeException (DecodeError str _) = T.pack str
+    unpackUnicodeException _                   = error "EncodeError is deprecated"
```
kutyel, line 178: 👍🏻

graphql/src/Mu/GraphQL/Server.hs
```diff
@@ -57,6 +56,7 @@ import           Network.Wai.Handler.Warp         (Port, Settings, run, runSetti
 import qualified Network.Wai.Handler.WebSockets   as WS
 import qualified Network.WebSockets               as WS
 
+import           Mu.GraphQL.Quasi.LostParser      (parseDoc)
```
kutyel, line 59: Is my "lost parser" still needed with this new parser? 🤔

kutyel (approved):
So much less `coerce`! 🥳

graphql/src/Mu/GraphQL/Annotations.hs
serras: This library requires a location for each node, so I'm creating a fake one. I'm planning to take advantage of these locations in error messages in the future, but now I just wanted to get this out.

graphql/src/Mu/GraphQL/Server.hs
serras: Now it's more like "function which runs the parser and juggles a bit with the result so it pleases what we need", but I decided to keep the name xD

PR #261, opened 2020-11-16
Fix GraphQL example + pools in mu-persistent

### LINKED ISSUE #259: Library example subscriptions: failed to resolve a field
Hello again!

I am using a `library` example.

This works:

```graphql
subscription {
  allBooks {
    title
  }
}
```

This doesn't:

```graphql
subscription {
  allBooks {
    title
    author {
      name
    }
  }
}
```

I have no idea how to fix resolving in subscriptions, so I need help.

serras:
Fixes #259

### DIFF
--- a/examples/library/src/Server.hs
+++ b/examples/library/src/Server.hs
@@ -109,31 +109,44 @@ libraryServer conn = resolver
     bookId :: Entity Book -> ServerErrorIO Integer
     bookId (Entity (BookKey k) _) = pure $ toInteger k
     bookTitle :: Entity Book -> ServerErrorIO T.Text
-    bookTitle (Entity _ Book { bookTitle = tl }) = pure tl
+    bookTitle (Entity _ Book { bookTitle }) = pure bookTitle
     bookAuthor :: Entity Book -> ServerErrorIO (Entity Author)
-    bookAuthor (Entity _ Book { bookAuthor = au }) = do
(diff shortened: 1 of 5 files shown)


serras:
The problem come from nested Persistent connections. I don't know how to fix that, so I've changed the code in `allBooksConduit` to bring all the elements in the table, and then create a conduit with them. It works.

examples/library/src/Server.hs
```diff
@@ -108,33 +109,37 @@ libraryServer conn = resolver
     bookId :: Entity Book -> ServerErrorIO Integer
     bookId (Entity (BookKey k) _) = pure $ toInteger k
     bookTitle :: Entity Book -> ServerErrorIO T.Text
-    bookTitle (Entity _ Book { bookTitle }) = pure bookTitle
+    bookTitle (Entity _ Book { bookTitle = tl }) = pure tl
```
kutyel, line 112: I think it looks simpler with `NamedFieldPuns`, what do you think? 😕

kutyel (approved):

examples/library/src/Server.hs
serras: 100% agree. I changed them because having unique names help me visualise where everything is coming from. I'll send a fix in a few minutes.

kutyel (approved):
🚀

PR #264, opened 2020-11-19
♻️ Run cabal-fmt and extract library example

kutyel:
@serras do you prefer this in 2 separate PRs?


serras:
The Nix file needs to be changed (and maybe the Stack ones too). I would prefer to keep building the backend as part of the general build, even though it also comes with its own `stack.yaml` for the cases in which this is seen separately.

serras:
Could be please not remove the backend from the Stack and Nix files, but simply point to `examples/library/backend`? I would really want to build this example every time we make any change, as it is the biggest one we have right now.

serras (approved):
Great job! 💯

PR #265, opened 2020-11-20
❄️  Add gitignore.nix to prevent Nix cache invalidation for trivial changes

kutyel:
The idea is that the gitignore.nix addition should prevent cache invalidation whenever something trivial changes in the repo 😉


serras:
Do you have any pointer about what the `pre-commit-hook` is doing? Having to run stylish-haskell is OK, but pleasing HLint is something I don't always want to do.

kutyel:
I generally find hlint tips helpful, that said they can be disabled `enabled = false` or we can simply add `HLINT IGNORE` annotations in our code whenever we don't want a hint 😉

serras:
I am not arguing about running it, I am just advocating for using the HLint action within our CI instead of a pre-commit hook.

serras (approved):

PR #260, opened 2020-11-15
Protobuf nested types

akshaymankar:
Fixes #258 

This is a breaking change, it adds scope (each level seprated by a ".") to names of all the nested types.

PR #266, opened 2020-11-22
Refer to nested types correctly in GRpc methods

akshaymankar:
I forgot to update these in #260


serras (approved):
Makes complete sense that also this needed changes. Thanks for your great work! :)

serras:
I've released a new version of `mu-protobuf` with this fix. Thanks again!

PR #268, opened 2020-11-30
Maps in Protocol Buffers

### LINKED ISSUE #267: Support for maps in protobuf
I see that they're [explicitly not supported](https://github.com/higherkindness/mu-haskell/blob/e99149be9f7a70e8c67c2b3ccd3ac1cd50e3606b/adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs#L356-L359) and print a type error. What would it take to support them? Do y'all have a preference between different map types which exist?

serras:
Fixes #267

### DIFF
--- a/adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
+++ b/adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
@@ -381,11 +381,16 @@ instance ( KnownNat (FindProtoBufId sch ty name)
     where fieldId = fromInteger $ natVal (Proxy @(FindProtoBufId sch ty name))
           go = PBDec.repeated (
                  PBDec.embedded'
-                   ((,) <$> PBDec.one protoToOneFieldValue
-                              (error "key should always be present") `at` 1
-                        <*> PBDec.one protoToOneFieldValue
-                              (error "value should always be present") `at` 2))
+                   ((,) <$> fieldValueWithDefault 1
+                        <*> fieldValueWithDefault 2))
                  `at` fieldId
+          fieldValueWithDefault innerFieldId
+            = case defaultOneFieldValue of
+                Nothing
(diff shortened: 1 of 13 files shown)


kutyel (dismissed):
LGTM, @akshaymankar could you take a look? 😉

adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs
```diff
@@ -353,10 +357,35 @@ instance {-# OVERLAPS #-}
...
+            = PBEnc.embedded fieldId $
+                oneFieldValueToProto 1 k <> oneFieldValueToProto 2 v
+  protoToField = Field . FMap . M.fromList <$> go
+    where fieldId = fromInteger $ natVal (Proxy @(FindProtoBufId sch ty name))
+          go = PBDec.repeated (
+                 PBDec.embedded'
+                   ((,) <$> PBDec.one protoToOneFieldValue
+                              (error "key should always be present") `at` 1
+                        <*> PBDec.one protoToOneFieldValue
+                              (error "value should always be present") `at` 2))
```
akshaymankar, line 387: It seems dangerous to throw error here. Maybe this can be handled by using `empty` like [here](https://github.com/higherkindness/mu-haskell/blob/aff0b446e7fce337e6158a918f4b24a489572972/adapter/protobuf/src/Mu/Adapter/ProtoBuf.hs#L317-L318)? 

If that is not possible for some reason, I think a parser error would be better. But it might be wrong, as according to the [docs](https://developers.google.com/protocol-buffers/docs/proto3#backwards_compatibility), this should be compatible with parsing the `MapFieldEntry` message type with key and value, and for that an old encoder may skip encoding an "empty" value in the map.

serras: You are completely right. I've fixed this in 67cc414, so now it behaves as any other field: if not present, it tries to use the default value, and otherwise the parsing itself fails.

kutyel (approved):
🔥

PR #280, opened 2021-01-13
Issue: example error parsing problem

jschaul:
This PR is twofold:

1. bug report
2. question about unit testing

## 1. bug report

As we deviate from the "happy path" (e.g. here when throwing an error), using mu-grpc-client becomes problematic.

To run this example:

```
stack exec error-parsing
```

Output:
```
testing things...
The result for saying hello to Alice: "GRpcOk \"hi, Alice\""
The result for saying hello to Bob: GRpcErrorString "not enough bytes"
done with the haskell test, waiting 10 seconds before shutting down...
```

As can be seen, any error thrown will be falsely interpreted by grpc-client as
"GRpcErrorString". This makes any real-life application with grpcclient
impossible (there are not just happy paths, and not all errors are the
same)

The server side implementation appears correct, as I can poke the server
with the grpcurl tool:

```
grpcurl -d '{"name": "Bobby"}' -format json -plaintext -proto errorparsing.proto localhost:8070 hello.Service.SayHello
{
  "reply": "hi, Bobby"
}

grpcurl -d '{"name": "Bob"}' -format json -plaintext -proto errorparsing.proto localhost:8070 hello.Service.SayHello
ERROR:
  Code: NotFound
  Message: Bob not there
```

Related issue: https://github.com/haskell-grpc-native/http2-grpc-haskell/issues/6

## Why am I filing this PR here?

Providing easy-to run example code along a github issue is hard, therefore I opted to use a PR instead.

More generally, I would have preferred to make this example a unit test (expecting the reply from Bob to be of a certain type which is not `GRpcErrorString "not enough bytes"`), however this whole project seems to not contain any unit tests at all.

## Question about unit tests and production-readyness

Do you plan to set up some tasty or hspec (or something else) unit tests more generally? (why are there no unit tests so far? That makes me doubt whether the mu family of libraries can safely be used in any project other than toy projects. Can you comment on the stability of the libraries outside this particular bug?) 

In case unit tests are more generally set up, this PR could be converted to be another unit test, which would allow to more easily catch other parsing issues in the mu family of libraries.

### DIFF
--- a/examples/error-parsing/src/Main.hs
+++ b/examples/error-parsing/src/Main.hs
@@ -1,0 +1,120 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveLift #-}
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE EmptyCase #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PackageImports #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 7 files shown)


kutyel:
Hey @jschaul thanks for opening this PR, looks good! 🙌🏻 

Regarding testing we have currently integration tests on a separate repo (see: https://github.com/higherkindness/mu-scala-haskell-integration-tests) that make sure that we communicate and work effectively with mu-scala, which is 100% production ready and used in other companies. Regarding unit tests it is true that @serras and I should think of a way of adding them, we didn't yet mainly because of lack of time and because since all the core parts of the library are hardcore type-level stuff... meaningful test coverage is hard to achieve 😢

examples/error-parsing/src/Main.hs
```diff
@@ -0,0 +1,120 @@
...
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE NoImplicitPrelude #-}
```
kutyel, line 44: Are these many lang extensions actually needed for this example? 🤯

examples/error-parsing/src/Main.hs
```diff
@@ -0,0 +1,120 @@
...
+runServer = do
+  runGRpcApp msgProtoBuf 8070 grpcServer
+
+------------------------------------------------------------------------------
+-- grpc server api
+
+grpcServer :: MonadServer m => SingleServerT i Service m _
+grpcServer =
+  singleService
+    ( method @"SayHello" sayHello
```
kutyel, line 76: I see you are a man of culture and are using ormolu 😉 👍🏻

examples/error-parsing/src/Main.hs
```diff
@@ -0,0 +1,120 @@
...
+-- test
+
+runTest :: IO ()
+runTest = do
+  putStrLn "testing things..."
+  aliceResult <- outboundSayHello' "127.0.0.1" 8070 "Alice"
+  putStr "The result for saying hello to Alice: "
+  print (show aliceResult)
+  bobResult <- outboundSayHello' "127.0.0.1" 8070 "Bob"
+  -- bobResult should give a valid error from the server (NotFound), instead of a generic "not enough bytes"
```
kutyel, line 111: The "not enough bytes" definitely seems like an error 🤔  we'll look into it!

PR #285, opened 2021-01-27
Use and require default value 0 for enums in Protobuf

### LINKED ISSUE #282: Protobuf: Enums have default values they shouldn't be required to be Maybe
If I have a protobuf like this:
```proto
enum MyEnum {
  Option1 = 0;
  Option2 = 1;
}

message MyMessage {
  MyEnum thing1 = 1;
  string thing2 = 2;
}
```

I am forced to create a data structure like this:

```haskell
data MyEnum = Option1 | Option2
data MyMessage = MyMessage { thing1 :: Maybe MyEnum, thing2 :: Text }
```

If I try to encode `MyMessage (Just Option1) "something"` and immediately decode it, I will get `MyMessage Nothing "something"`. This is because enums default to the first value and so in protobuf they are encoded the same as being not specified.

I think the `protobuf` and `grpc` TH functions should make the schema in a way that enums are not marked as optional.

serras:
Fixes #282 

@akshaymankar @jschaul is this enough to remove the workaround from your code?

### DIFF
--- a/adapter/avro/test/Avro.hs
+++ b/adapter/avro/test/Avro.hs
@@ -20,9 +20,9 @@ exampleAddress = Address "1111BB" "Spain"
 
 examplePerson1, examplePerson2 :: Person
 examplePerson1
-  = Person "Haskellio" "Gomez" (Just 30) (Just Male) exampleAddress [1,2,3] M.empty
+  = Person "Haskellio" "Gomez" (Just 30) Male exampleAddress [1,2,3] M.empty
 examplePerson2
-  = Person "Cuarenta" "Siete" Nothing Nothing exampleAddress []
+  = Person "Cuarenta" "Siete" Nothing Unspecified exampleAddress []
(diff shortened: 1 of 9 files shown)


kutyel (approved):
🔥

serras:
Now it's really ready, I forgot to change the Template Haskell so it was still generating `TOption`. Now it handles enums correctly.

akshaymankar:
This works as expected! Thanks for the fix :smile:

serras:
`mu-protobuf` version 0.4.2.0, which includes this fix, has been released to Hackage

akshaymankar:
@serras The change in `core/schema/src/Mu/Schema/Examples.hs` is required for mu-protobuf to be able to compile, can you please release that to hackage too?

serras:
Done! `mu-schema-0.3.1.2` has been released.

Issue #293, opened 2021-03-29
Support for arbitrary scalars? (GraphQL)


lfborjas:
Hi there. Thank you very much for this excellent collection of libraries! Watching Alejandro's videos about the graphql capabilities inspired me to write my latest service in GraphQL, using `mu-graphql`! 

However, I'm stuck on something:

In my schema, I'm planning to include a `DateTime` scalar that I would like to map to a `LocalTime` in Haskell. As [per the graphql spec](https://spec.graphql.org/June2018/#sec-Scalars), here's an example schema:

```grapqhl
scalar DateTime

type User{
  dob: DateTime!
}

type Query{
  firstByDoB(dob: DateTime!): User
}
```

I'm not quite sure what to tell the Haskell side, however; I tried adding a naïve mapping:

```haskell
type ServiceMapping = '[
  , "DateTime" ':-> LocalDateTime
  ]
```

Where `LocalDateTime` is a `newtype` over `LocalTime` (from `Data.Time`).  If I omit the mapping, the generated type seems to infer that the type of the `DateTime` argument in the query is `()`.

The crux of the matter is: I don't quite know how to wire it up: what to tell my `resolver` so it can actually act on that mapping: should I use `method` at the top level? That doesn't seem right, since it's a type, not a query/mutation -- and If I do that, I get `could not find method "DateTime"` errors, as well as:  `No instance for (mu-graphql-0.5.0.2:Mu.GraphQL.Query.Parse.ParseArg`, anyway 😅 
  
If I try a wild guess and derive schema instances manually, while keeping the mapping:

```haskell
newtype LocalDateTime 
  = LocalDateTime LocalTime
  deriving newtype (Eq, Ord, Show)
  deriving (Generic)

instance FromSchema MyApiSchema "DateTime" LocalDateTime
instance ToSchema MyApiSchema "DateTime" LocalDateTime
```

I get "cannot find type 'DateTime' in the schema" errors in addition to the previous ones, which seems to point to it not being recognized, maybe? But at this point, I'm just flailing: it's just not clear to me how an arbitrary scalar is supposed to be "wired into" the mechanism, short of writing the schema by hand? I'm happy to write the instances of `FromSchema` and `ToSchema` by hand, since it would make sense to be responsible for how the Strings that are sent over the wire egress/ingress Haskell-land, but I think I'm stuck at an earlier stage than that.

[...]

PR #295, opened 2021-04-01
Add support for new primitive types in GraphQL

serras:
This is a first attempt at solving #293. The idea is that information about more primitive types would be injected as:

```haskell
graphqlWithExtendedPrimitives [("DateTime", [t|LocalDateTime|])] ...
```


kutyel (dismissed):
LGTM, while we fix this failing branch can we also somehow fix master as well when we deploy? 🤣  (I don't know what fails when we merge to master 🤔 )

kutyel (approved):
🔥

PR #297, opened 2021-04-22
[protobuf] Fix parsing of oneof fields

akshaymankar:
In protobuf, unions themselves are optional, so while parsing if any value fails
to parse, its default must not be considered. When the default is considered and
first value in the union is not specified, the union still gets parsed as the
default of the first value. So, in this instance:

```proto
oneof foo {
  int32 foo_int = 1;
  string foo_str = 2;
}
```

`foo_str` will never be parsed, as foo_int will have a default, i.e. 0 and so
`foo_str "blah"` will always wrongly parse as `foo_int 0`.

/cc @pcapriotti


kutyel (approved):
Thanks for your contribution! 🙌🏻

serras:
Thanks for your contribution, @akshaymankar. We are not experts in Protocol Buffers, and that protocol seems to have quite some corner cases!

One question I have is why you changed the `test`. I see that now the protobuf file is imported (instead of manually defined), but I also see some changes in the handling of `Foo` that I do not completely understand.

akshaymankar:
> One question I have is why you changed the `test`. I see that now the protobuf file is imported (instead of manually defined), but I also see some changes in the handling of `Foo` that I do not completely understand.

Yeah, I changed the test because the schema defined in haskell did not line up with the definitions in the proto file, importing it just ensured that they're at least always compatible. Can you please elaborate on what you don't understand about handling of `Foo`, I can try to explain.

serras:
> Yeah, I changed the test because the schema defined in haskell did not line up with the definitions in the proto file, importing it just ensured that they're at least always compatible. 

That makes complete sense.

> Can you please elaborate on what you don't understand about handling of `Foo`, I can try to explain.

I saw you included hand-written instances of `FromSchema` in the test, and I was wondering why that was the case.

akshaymankar:
> I saw you included hand-written instances of `FromSchema` in the test, and I was wondering why that was the case.

Ah yes, this is because, as far as I understand, writing `ToSchema` and `FromSchema` for `oneof` fields is not possible. The example here is too simple to see this and I made it so to isolate the parts which need me to write the instances manually. To understand this, take this example:

```proto
message Person {
  string name = 1;
  oneof vehicle {
    Bicycle bicycle = 2;
    Motorbike motorbike  3; 
  }
}

message Bicycle {
  string color = 1;
}

message Motorbike {
  string registration_number = 1;
}
```

Here, the corresponding haskell type would look something like this:

```haskell
data Person = Person { name :: Text, vehicle :: Maybe Vehicle }
data Vehicle = VBicycle Bicycle | VMotorbike MotorBike
data Bicycle = Bicycle { color :: Text }
data Motorbike = Motorbike { registration_number :: Text }
```

And the corresponding mu-schema looks like this:

```haskell
type ExampleSchema =
    '[ 'DRecord "Person" '[ 'FieldDef "name" ('TPrimitive T.Text),
                            'FieldDef "vehicle" ('TUnion '[ 'TSchematic "Bicycle",
                                                            'TSchematic "Motorbike"])],
       'DRecord "Bicycle" '[ 'FieldDef "color" ('TPrimitive T.Text)],
       'DRecord "Motorbike" '[ 'FieldDef "registration_number" ('TPrimitive T.Text)]]
type instance AnnotatedSchema ProtoBufAnnotation ExampleSchema = '[ 'AnnField "Person" "name" ('ProtoBufId 1 '[]),
                                                                    'AnnField "Person" "vehicle" ('ProtoBufOneOfIds '[2,
                                                                                                                      3]),
                                                                    'AnnField "Bicycle" "color" ('ProtoBufId 1 '[]),
                                                                    'AnnField "Motorbike" "registration_number" ('ProtoBufId 1 '[])]
```

Notice that the field `vehicle` is a `TUnion` here and so has no name. So, if I want to write an instance for `ToSchema ExampleSchema <name> Vehicle`, I cannot. This forces me to hand write `ToSchema ExampleSchema "Person" Person`, to avoid this I created `Foo` as a separate message and wrote a smaller instances by hand. That said, I think it is be possible to derive the instances if the haskell types as defined like this:

```haskell
data Person = Person { name :: Text, vehicle :: Maybe (NS I [Bicycle, Motorbike]) }
data Bicycle = Bicycle { color :: Text }
data Motorbike = Motorbike { registration_number :: Text }
```

But, this will require a generic based instances of `ToSchema` and `FromSchema` for `NS` which don't exist. Maybe this can be a separate issue to work on.

serras:
@akshaymankar I added some changes to the PR, but did them in #302, because I had no permissions to push in your branch

akshaymankar:
> @akshaymankar I added some changes to the PR, but did them in #302, because I had no permissions to push in your branch

Sounds good, please feel free to merge it :+1:

PR #306, opened 2021-04-30
Do not create a loop in bidi gRPC

### LINKED ISSUE #300: Bi-Directional streaming consumes tons of CPU
I found that a server consumes around 100% of CPU load during bi-directional RPC.

I can repro the phenomenon even in a very simple call like the followings. On the other hand, libraries in other languages (e.g. grpc-go) don't that. So I'm afraid it is from an immanent problem in Mu-Haskell implementation. How can I tame it?

```haskell
import           Data.Conduit
import qualified Data.Conduit.Combinators as C

sayManyHellos
  :: (MonadServer m)
  => ConduitT () HelloRequest m () -> ConduitT HelloResponse Void m () -> m ()
sayManyHellos src sink = do
    let reply (HelloRequest n) = HelloResponse ("Hello, " <> n)
    runConduit $ src .| C.map reply .| sink
```

```
syntax = "proto3";

package quickstart;

service Greeter {
  rpc SayManyHellos (stream HelloRequest) returns (stream HelloResponse) {}
}

message HelloRequest { string name = 1; }
message HelloResponse { string message = 1; }
```

```haskell
grpc "TheSchema" id "greeter.proto"

data HelloRequest = HelloRequest {
      name :: T.Text
    } deriving (
      Eq, Show, Generic
    , ToSchema   TheSchema "HelloRequest"
    , FromSchema TheSchema "HelloRequest"
    )

data HelloResponse = HelloResponse {
      message :: T.Text
    } deriving (
      Eq, Show, Generic
    , ToSchema   TheSchema "HelloResponse"
    , FromSchema TheSchema "HelloResponse"
    )
```

serras:
Fixes #300

@y-taka-23 could you check that this fixes the problem you've reported?


grpc/server/src/Mu/GRpc/Server.hs
```diff
@@ -465,15 +465,15 @@ instance (GRpcInputWrapper p vref v, GRpcOutputWrapper p rref r, MonadIO m)
...
+                         Just o ->
                            pure $ Just ((), buildGRpcOWTy (Proxy @p) (Proxy @rref) o)
-                         Just Nothing  -> do
+                         Nothing -> do
                            liftIO $ cancel promise
                            pure Nothing
-                         Nothing -> -- no new elements to output
-                           readNext ()
+                         -- Nothing -> -- no new elements to output
+                         --   readNext ()
```
kutyel, line 476: Should we remove the comments? 🤔

kutyel (approved):
LGTM

grpc/server/src/Mu/GRpc/Server.hs
serras: Once we know this fix solves the issue, I'll do some beautification of the code.

y-taka-23:
@serras Thank you! Looks nicely fixed as tasting with the `stack.yaml`:

```yaml
extra-deps:
  - mu-rpc-0.5.0.0
  - mu-grpc-common-0.4.0.0
  - github: higherkindness/mu-haskell
    commit: cc91a1767f8517919437542ab6d3df9a78adff8d
    subdirs:
      - grpc/server
  (snip)
```

PR #309, opened 2021-04-30
Multiplex gRPC servers

### LINKED ISSUE #244: Support multiplexing many GRPC servers on same port
Many implementations of GRPC support this trivially, but it doesn't seem to be possible with mu-grpc-server.

Here is how it is done in golang (Example copied from https://github.com/grpc/grpc/issues/4418)):

```golang
    srv := grpc.NewServer()
    rpx.RegisterBooksServer(srv, booksSvc)
    rpx.RegisterMoviesServer(srv, mediaSvc)
    rpx.RegisterGamesServer(srv, mediaSvc)
```
I thnk this can be done if [`Mu.Server.ServerT`](https://github.com/higherkindness/mu-haskell/blob/5d9b01ddaa6aa7212c0906327e1e14b012046742/core/rpc/src/Mu/Server.hs#L140-L144) could be defined for a list of packages instead of just one package.

serras:
Closes #244

### DIFF
--- a/grpc/server/src/Mu/GRpc/Server.hs
+++ b/grpc/server/src/Mu/GRpc/Server.hs
@@ -29,7 +29,7 @@ module Mu.GRpc.Server
 , runGRpcAppTLS, TLSSettings
   -- * Convert a 'Server' into a WAI application
 , gRpcApp, gRpcAppTrans
-, gRpcMultipleApp, gRpcMultipleAppTrans
+, WrappedServer(..), gRpcMultipleApp, gRpcMultipleAppTrans
   -- * Raise errors as exceptions in IO
 , raiseErrors, liftServerConduit
   -- * Re-export useful instances
@@ -154,9 +154,8 @@ gRpcAppTrans protocol f svr
 --   for example, @wai-routes@, or you can add middleware
 --   from @wai-extra@, among others.
 gRpcMultipleApp
-  :: ( ToMultipleServers protocol ServerErrorIO srvs )
-  => Proxy protocol
-  -> srvs
+  :: Proxy protocol
+  -> [WrappedServer protocol ServerErrorIO]
   -> Application
 gRpcMultipleApp protocol = gRpcMultipleAppTrans protocol id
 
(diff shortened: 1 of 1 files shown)


akshaymankar:
For my use case at least, it would be more ergonomic if `ToMultipleServers` was also defined for a type level list as I want to be able to merge arbitrarily many servers. 
Also, is it ok if `MultipleServers` and `ToMultipleServers` are exposed?

akshaymankar:
On closer inspection, `MultipleServers` is the "type level list" I wanted to make. What do you think about exposing that and instead of relying on `ToMutipleServers` type class? The user of the library can be expected to build `MultipleServers` using the constructors `MSEnd` and `MSOneMore`.

grpc/server/src/Mu/GRpc/Server.hs
```diff
@@ -29,6 +29,7 @@ module Mu.GRpc.Server
 , runGRpcAppTLS, TLSSettings
   -- * Convert a 'Server' into a WAI application
 , gRpcApp, gRpcAppTrans
+, gRpcMultipleApp, gRpcMultipleAppTrans
```
kutyel, line 32: As per @akshaymankar suggestion, we should consider exporting `MultipleServers` too?
```suggestion
, gRpcMultipleApp, gRpcMultipleAppTrans, MultipleServers(..)
```

kutyel (dismissed):

grpc/server/src/Mu/GRpc/Server.hs
akshaymankar: Also `gRpcMultipleServerHandlers`. Also, if that is being exposed it might be best to remove `ToMultipleServers` and instead provide functions to just translate tuples into `MultipleServers` and make these functions directly depend on `MultipleServers`

serras:
I've pushed a simplification of the approach: instead of using a hand-rolled heterogeneus list (`MultipleServers`), we can simply wrap each element using a GADT. As a result, the API uses a normal list, which I find much easier to handle:

```haskell
gRpcMultipleApp msgAvro [Srv server1, Srv server2]
```

@akshaymankar @kutyel thoughts? If that seems OK, I'll add some documentation and merge the PR.

kutyel (approved):
LGTM!

akshaymankar:
Looks good to me too :100:

PR #312, opened 2021-05-28
[grpc-server] Close bidirectional streaming requests correctly

### LINKED ISSUE #311: [grpc-server] Bidirectional streaming is not ended correctly
Testing a simple service defined like this:

```proto
syntax = "proto3";
package mu.grpc.test;

message GreetingRequest {string name = 1; }

message GreetingResponse {string greeting = 1;}

service GreetingsService {
  rpc bidirectional (stream GreetingRequest) returns (stream GreetingResponse);
}
```

The bidirectional function is implemented like this:

```haskell
bidirectional :: forall m. (MonadServer m, MonadIO m) => ConduitT () GreetingRequest m () -> ConduitT GreetingResponse Void m () -> m ()
bidirectional requests responses = do
  C.runConduit $ requests .| serveResponse .| responses
  liftIO $ putStrLn "Done!"
  where
    serveResponse :: ConduitT GreetingRequest GreetingResponse m ()
    serveResponse = do
      liftIO $ putStrLn "Server Waiting"
      mreq <- C.await
      case mreq of
        Nothing -> pure ()
        Just req -> do
          C.yield $ record1 $ "Greetings! " <> view #name req
          serveResponse
```

While testing with [evans](https://github.com/ktr0731/evans#bidirectional-streaming-rpc), I get this error at the end:

```
command call: failed to process bidi streaming RPC: rpc error: code = Internal desc = AsyncCancelled
```

Another thing to note is that `"Done!"` is never printed. So, it looks like the `requests` conduit never realises that the input is over.

akshaymankar:
Changes:
- Don't cancel the handler when server stream ends, this will allow the handler
to finish.
- Use TMChan instead of TMVar for consuming the server stream, this allows the
handler to indicate that it has finished and there will be no more output being
produced.

Fixes #311

### DIFF
--- a/grpc/server/src/Mu/GRpc/Server.hs
+++ b/grpc/server/src/Mu/GRpc/Server.hs
@@ -492,29 +492,34 @@ instance (GRpcInputWrapper p vref v, GRpcOutputWrapper p rref r, MonadIO m)
                    -> m ( (), IncomingStream m (GRpcIWTy p vref v) ()
                         , (), OutgoingStream m (GRpcOWTy p rref r) () )
           bdstream req = do
-            -- Create a new TMChan and a new variable
-            chan <- liftIO newTMChanIO :: m (TMChan v)
-            let producer = sourceTMChan @m chan
-            var <- liftIO newEmptyTMVarIO :: m (TMVar (Maybe r))
+            -- Create a new TMChan for consuming the client stream, it will be
+            -- the producer for the conduit.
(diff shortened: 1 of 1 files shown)


kutyel (approved):
Thanks! Looks great! 👀  💯  Let's wait for an additional review from @serras before merging 😉

serras (approved):
Looks good. Thanks for the hard work!

